PR #47, opened 2015-01-14
Pin Beaker-rspec to 4.x until fixed

cyberious:


Gemfile
```diff
@@ -25,13 +25,9 @@ group :development, :test do
...
   gem 'pry',                     :require => false
+  gem 'pry-byebug',              :require => false
   gem 'simplecov',               :require => false
-  if beaker_version
-    gem 'beaker', *location_for(beaker_version)
-  else
-    gem 'beaker',                :require => false, :platforms => :ruby
-  end
-  gem 'beaker-rspec',            :require => false, :platforms => :ruby
+  gem 'beaker-rspec', *location_for(beaker_rspec_version)
```
ferventcoder: Does this mean you don't even use beaker or that beaker-rspec takes a dependency on beaker as well?

Gemfile
```diff
@@ -25,13 +25,9 @@ group :development, :test do
   gem 'serverspec',              :require => false
   gem 'puppet-lint',             :require => false
   gem 'pry',                     :require => false
+  gem 'pry-byebug',              :require => false
```
ferventcoder: Typically used for ruby 2 only. Most folks create a Gemfile.local with this stuff in it. I have a selection based on which version of ruby I'm using but I guess it doesn't make a difference in about a month, eh?

cyberious: Yeah should not have been added will update once I get to office

Gemfile
cyberious: Beaker is a dependency of beaker-rspec

Gemfile
```diff
@@ -14,24 +14,16 @@ gem "puppet", *location_for(ENV['PUPPET_LOCATION'] || '~> 3.4.0')
...
 group :development, :test do
   gem 'rspec'
   gem 'mocha'
   gem 'mime-types', '<2.0',      :require => false
   gem 'rake',                    :require => false
-  gem 'rspec-puppet',            :require => false
   gem 'puppetlabs_spec_helper',  :require => false
-  gem 'serverspec',              :require => false
   gem 'puppet-lint',             :require => false
-  gem 'pry',                     :require => false
```
ferventcoder: Ha, didn't mean you needed to remove pry. See you are removing other things as well. `serverspec`, `rspec-puppet` - was that intentional?

cyberious: It is, they are either not needed or required by beaker-rspec so they will be pulled in.

PR #49, opened 2015-03-24
(BKR-147) add Gemfile setting for BEAKER_VERSION for puppet...

anodelman:
puppetdb, etc

- support for BEAKER_VERSION and BEAKER_RSPEC_VERSION in gemfile

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -14,6 +14,7 @@ gem "puppet", *location_for(ENV['PUPPET_LOCATION'] || '~> 3.4.0')
 gem "facter", *location_for(ENV['FACTER_LOCATION'] || '~> 1.6')
 gem "hiera", *location_for(ENV['HIERA_LOCATION'] || '~> 1.0')
 
+beaker_version = ENV['BEAKER_VERSION']
 beaker_rspec_version = ENV['BEAKER_RSPEC_VERSION'] || '~> 4.0'
 group :development, :test do
   gem 'rspec'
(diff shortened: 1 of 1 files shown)


Gemfile
```diff
@@ -23,7 +24,14 @@ group :development, :test do
...
   gem 'simplecov',               :require => false
-  gem 'beaker-rspec', *location_for(beaker_rspec_version)
+  if beaker_version
+    gem 'beaker', *location_for(beaker_version)
+  end
+  if beaker_rspec_version
+    gem 'beaker-rspec', *location_for(beaker_rspec_version)
+  else
+    gem 'beaker-rspec',  :require => false
+  end
```
ferventcoder: Does beaker-rspec work on Windows? If no, this will need further shielding

justinstoller: This PR doesn't change the exposure of Windows to beaker-rspec. ie it was installed before, and installation will still be attempted after this PR, it will just use more of Bundler's options when installing it. If the Gemfile didn't work before, it still won't work, if it worked before, it should continue to work.

anodelman: Yes, @justinstoller is correct.  This doesn't alter the functionality/defaults of the Gemfile - just adds support for BEAKER_VERSION/BEAKER_RSPEC_VERSION.

ferventcoder: Let's keep it broken is not really a good statement guys. It's too easy to add `:systemtests` around this - it will make it consistent with other modules as well.

anodelman: Would you prefer to simply remove the beaker/beaker-rspec gem dependency until such time as there are acceptance level tests?  If so, I'm cool with that.  I was just folding the Gemfile under the new standard without really being concerned with the actual contents of that file.

ferventcoder: Actually, we can take this as is - let's just file something or followup with a fix.

ferventcoder: I'm not sure what the new standard is, I just saw that other gemfiles use `:system_tests` and thought that would be part of this change.

PR #50, opened 2015-03-25
Edits to grammar, markdown, and format to match current styleguide.

jtappa:
Per DOC-1497

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,7 @@
 1. [Overview](#overview)
 2. [Module Description - What the module does and why it is useful](#module-description)
 3. [Setup - The basics of getting started with powershell](#setup)
+    * [Setup requirements](#setup-requirements)
     * [Beginning with powershell](#beginning-with-powershell)
 4. [Usage - Configuration options and additional functionality](#usage)
 5. [Reference - An under-the-hood peek at what the module is doing and how](#reference)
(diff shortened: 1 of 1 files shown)


ferventcoder:
Where should we document that reorder is a different commit at the end? Make it easy to see what changed in the first commit. I had this conversation already with the acls doc fixups

jbondpdx:
So, @ferventcoder , we had a team discussion about this issue, and we don't think it's a good practice for us to separate commits in that way. 

The biggest reason for that is that these PRs also serve as a technical review, which means the document really needs to be reviewed as a whole, rather than just what's been changed. We've been bitten before when only the changes have gotten reviewed, so we'd like to avoid that in the future.

ferventcoder:
@jbondpdx what part of the technical review as a whole cannot be reviewed? I guess not including me in your discussion you missed that you can do BOTH if you split the commits.

ferventcoder:
Let me explain. When you split the commits up, I can go to the first commit and look at exactly what has changed without the second commit where things were reordered. OR I can click over to Files changed tab and see the entire set of changes as a whole and comment on them as well. It's a win win for everyone. What @psoloway is doing in that other commit wasn't exactly as I had intended when I asked for the changes to be split. I had asked that the FIRST commit be the changes to the doc, and the second commit be the reorder. 

With both changes up in the PR, you can get BOTH things you are looking for.

jbondpdx:
Ah, I see what you're saying. And yes, I think your request was somewhat misunderstood. 

That said, I'm still not sure why they need to be separate commits if you're reading the entire document for correctness. We need devs to review what *hasn't* been changed as well as what has, so I'm still not seeing why it is important for the changes to be separated in this way.

That said, we also don't want to do this because it's rather onerous. Reordering and editing aren't separate tasks; they're interdependent. So separating these wouldn't be a win for us at all, but would instead create an undue burden.

ferventcoder:
> That said, I'm still not sure why they need to be separate commits if you're reading the entire document for correctness. We need devs to review what hasn't been changed as well as what has, so I'm still not seeing why it is important for the changes to be separated in this way.

It's hard to see what HAS changed when the order is different. If I can go to a specific commit and look at what has changed, and then go to the changes part of the PR and see it all together I can see the changes as a whole.

The important part in some cases is seeing that the meaning of the changes hasn't been lost, which is not easy to see when the order is changed and the words are changed in the ***same*** commit. 

See https://github.com/puppetlabs/puppetlabs-acl/pull/53/files#diff-04c6e90faac2675aa89e2176d2eec7d8R17 for an example of seeing what has changed.

Now see https://github.com/puppetlabs/puppetlabs-acl/pull/53/files#diff-04c6e90faac2675aa89e2176d2eec7d8R53 for an OMG I have no idea if this is reordered or changed or both. If there are two distinct commits, I can go back to the first commit and look at what specifically changed. 

> That said, we also don't want to do this because it's rather onerous. Reordering and editing aren't separate tasks; they're interdependent. So separating these wouldn't be a win for us at all, but would instead create an undue burden.

I can totally understand what you mean here because it creates an undue burden on the reviewer as it's nearly impossible to see what has changed. Maybe there is a happy medium here?

ferventcoder:
That all said, I love the help making the documentation better. I definitely think there is a happy medium we can achieve that keeps the burden reasonable for both sides.

ferventcoder:
I put those links in using a split view, if you are using unified you may want to scroll up to see what the diff is comparing.

ferventcoder:
Perhaps the first commit is the reorder and then subsequent commits are related to getting the wording right? That seems to be a more plausible method of going about it. Plus there are methods to compare the wording changes based on this methodology (which will avoid interactive rebases).

jbondpdx:
Are we good to merge these changes?

ferventcoder:
@jbondpdx @jtappa The commit message needs fixed to indicate DOC-1497. I did my best on a review of the actual changes. 

Please understand that my suggestions above were to help alleviate more work on your part as well in telling me what you changed. Without following good git practices, I may need more hand-holding with regard to the actual wording changes.

README.md
```diff
@@ -17,95 +18,110 @@ This module adds a new exec provider capable of executing PowerShell commands.
 
 ##Module Description
 
-Puppet provides a built-in `exec` type that is capable of executing commands. This module adds a `powershell` provider to the `exec` type,  which enables all of the `exec` parameters, such as `creates`, `onlyif`, `unless`, etc. This module is particularly helpful if you need to run PowerShell commands but don't know the details about how PowerShell is executed, since you can technically run PowerShell commands in Puppet without the module.
+Puppet provides a built-in `exec` type that is capable of executing commands. This module adds a `powershell` provider to the `exec` type,  which enables `exec` parameters, listed below. This module is particularly helpful if you need to run PowerShell commands but don't know the details about how PowerShell is executed, since you can technically run PowerShell commands in Puppet without the module.
```
ferventcoder: Should we link to the parameters section in the `listed below`?

README.md
```diff
@@ -17,95 +18,110 @@ This module adds a new exec provider capable of executing PowerShell commands.
...
     }
-```
+~~~
 
 ##Usage
 
 When using `exec` resources with the `powershell` provider, the `command` parameter must be single-quoted to prevent Puppet from interpolating `$(..)`. 
 
-For instance, if you wanted to rename the Guest account
+For instance, if you wanted to rename the guest account:
```
ferventcoder: The `Guest` account is capital. That's the name of the account.

README.md
```diff
@@ -17,95 +18,110 @@ This module adds a new exec provider capable of executing PowerShell commands.
...
       command   => '$(Get-WMIObject Win32_UserAccount -Filter "Name=\'guest\'").Rename("new-guest")',
       unless    => 'if (Get-WmiObject Win32_UserAccount -Filter "Name=\'guest\'") { exit 1 }',
       provider  => powershell,
     }
-```
+~~~
 
-Note that the example uses the `unless` parameter to make the resource idempotent. The `command` is only executed if the Guest account does not exist, as indicated by
-`unless` returning 0.
+Note that the example uses the `unless` parameter to make the resource idempotent. The `command` is only executed if the guest account does not exist, as indicated by `unless` returning 0.
```
ferventcoder: Again, `Guest` not `guest`.

README.md
```diff
@@ -17,95 +18,110 @@ This module adds a new exec provider capable of executing PowerShell commands.
...
 $obj = $(Get-WMIObject Win32_UserAccount -Filter "Name='Guest'")
 $obj.Rename("OtherGuest")
-```
+~~~
 
 This has the added benefit of not requiring escaping '$' in the PowerShell code.
 
 ##Reference
 
-Since the powershell module is a provider for `exec` it enables the same parameters as [`exec`](http://docs.puppetlabs.com/references/stable/type.html#exec) for PowerShell.
```
ferventcoder: Did this wording go away or get moved somewhere else?

ferventcoder: It's helpful to have the link. Not necessary though.

README.md
```diff
@@ -17,95 +18,110 @@ This module adds a new exec provider capable of executing PowerShell commands.
...
-* **refresh** - How to refresh the command.
-* **refreshonly** -  Used with `subscribe` and `notify` [metaparameters](http://docs.puppetlabs.com/references/latest/metaparameter.html) to refresh the command only when a dependent object is changed. Valid values are 'true' and 'false'.
-* **returns** - The expected return code(s). An error will be returned if the executed command returns something else. Defaults to 0. Can be specified as an array of acceptable return codes or a single value.
-* **timeout** - The maximum time the command should take.
-* **tries** - The number of times execution of the command should be tried. Defaults to '1'. 
-* **try_sleep** - The time to sleep in seconds between `tries`.
-* **umask** - The umask to be used while executing the command.
-* **unless** - The `exec` will run unless the command returns 0.
+#####`command`
+Specifies the actual PowerShell command to execute. Must either be fully qualified or a search path for the command must be provided. Valid options: String. Default: Undefined.
```
ferventcoder: `String` is confusing to me.

README.md
ferventcoder: Actually because we are removing a bunch of parameters from below, it would be most helpful to have this link re-added. And then parameters should state it builds upon the existing parameters/properties already available in the exec resource.

README.md
```diff
@@ -17,95 +18,110 @@ This module adds a new exec provider capable of executing PowerShell commands.
...
 
-##Limitations
+#####`environment`
+Sets additional environment variables to set for a command. Valid options: String, or an array of multiple options. Default: Undefined.
```
ferventcoder: `String` is weird. The wording is telling me a type I can use, but nearly everything is a string in puppet 3 and below.

README.md
ferventcoder: Missing options:
 * `creates`
 * `cwd`
 * `group`
 * `unmask`

If those don't apply to this provider, perhaps we should mention that somewhere so there is not confusing in trying to use them. Thoughts?

README.md
```diff
@@ -17,95 +18,110 @@ This module adds a new exec provider capable of executing PowerShell commands.
...
- * Only supported on Windows Server 2003 and above, and Windows 7 and 8
+#####`logoutput`
+Defines whether to log command output in addition to logging the exit code. If you specify 'on_failure', it only logs the output when the command has an exit code that does not match any value specified by the `returns` attribute. Valid options: 'true', 'false', and 'on_failure'. Default: 'on_failure'.
 
-##Development
+#####`onlyif`
+Runs the exec only if the command returns 0. Valid options: String. Default: Undefined.
+
+#####`path`
+Specifies the search path used for command execution. Valid options: String of the path, or a ; separated list. Default: Undefined.
```
ferventcoder: Semi-colon separated list may be better spelled out. I've also seen this as an array (not sure if it works or not).

Reference where I saw it most recently - https://github.com/puppetlabs/puppetlabs-sqlserver/pull/83#discussion_r27883091

README.md
```diff
@@ -17,95 +18,110 @@ This module adds a new exec provider capable of executing PowerShell commands.
...
+#####`refreshonly`
+Refreshes the command only when a dependent object is changed. Used with `subscribe` and `notify` [metaparameters](http://docs.puppetlabs.com/references/latest/metaparameter.html). Valid options: 'true', 'false'. Default: 'false'.
 
-Puppet Labs modules on the Puppet Forge are open projects, and community contributions are essential for keeping them great. We can’t access the huge number of platforms and myriad of hardware, software, and deployment configurations that Puppet is intended to serve.
+#####`returns`
+Lists the expected return code(s). An error will be returned if the executed command returns something else. Valid options: An array of acceptable return codes or a single value. Default: 0.
 
-We want to keep it as easy as possible to contribute changes so that our modules work in your environment. There are a few guidelines that we need contributors to follow so that we can have a chance of keeping on top of things.
+#####`timeout`
+Sets the maximum time in seconds that the command should take. Valid options: Number or string representation of a number. Default: 300.
```
ferventcoder: :+1:  on getting the default specified

README.md
```diff
@@ -17,95 +18,110 @@ This module adds a new exec provider capable of executing PowerShell commands.
...
+ 
+#####`try_sleep`
+Specifies the time to sleep in seconds between `tries`. Valid options: Number or a string representation of a number. Default: Undefined.
+
+#####`unless`
+Runs the `exec`, unless the command returns 0. Valid options: String. Default: Undefined.
+
+##Limitations
+
+ * Only supported on Windows Server 2003 and above, and Windows 7 and 8.
```
ferventcoder: I would state the second part as Windows 7+. That will save updating documentation when Windows 10 comes out.

ferventcoder: Or Windows 7 and above (like the former statement).

jtappa: Is it safe to assume though that it will automatically be supported in newer versions of Windows?

README.md
jtappa: There were previously 3 mentions of this module using exec's params in the readme, so to cut down on cruft, it was removed here. Our readme style guide also tries to avoid links as much as possible, as many of our users are reading this readme in vim or offline and markdown links make the document harder to read. The parameters that were removed from the readme were deemed not relevant to the powershell module, as previously the list of params was just copied from exec's documentation.

README.md
jtappa: There's already a link in the ToC to the reference, so a link is not necessary here. We're trying to avoid excessive linking in the readmes. The "listed below" was added in conjunction with the removal of the link to exec's documentation to help clarify we're specifically listing relevant parameters in the readme.

README.md
jtappa: What would you suggest instead of string?

README.md
jtappa: :+1:

README.md
jtappa: :+1:

README.md
jtappa: If you've seen it as an array, and you'd like that listed as a valid option, could you please verify that it works?

README.md
ferventcoder: :+1:

README.md
ferventcoder: Its hard to say if there is value in keeping them for the reasons you just mentioned - in vim or no internet. I would default to keeping them - they are valid parameters. Thoughts? Cc @joshcooper

README.md
ferventcoder: I don't have a great suggestion yet. Just that string is a type and not a value. But I get that you changed that to valid "options", not values.

README.md
ferventcoder: I can give it a shot sure

README.md
ferventcoder: Yes

ferventcoder: However we might want to separate out where it is known to work versus what versions of Windows are tested and fall under the "supported module" aspect. I might be surprised to learn that we may not support the desktop versions of Windows. I'm not saying that's the case here. :)

README.md
joshcooper: Hey @jtappa, `creates` is important, because it can be used (like `onlyif` and `unless`) to make the powershell resource idempotent. The `cwd` is somewhat important, depending on the use case. I agree `group` and `umask` are not (they're meaningless for this provider).

README.md
cyberious: https://docs.puppetlabs.com/references/latest/type.html#exec-attribute-path we have the exec provider documented as an array, but can also be a ; on windows or : on *nix respectively.  Specifically the refreshonly has the example of path => [...] if this is not correct than we need to change this doc as well, but my experience is that it does work.

README.md
jtappa: I'll re-add `creates` and `cwd`. What are the defaults for these parameters, if any?

README.md
jtappa: Thanks Travis for the confirmation. I'll add array as an option here.

README.md
jtappa: For now, I'll just change this sentence to end in "Windows 7 and above."  
If we have an official list of tested/working versions, it can be included in the future but for now I'd like to get these changes merged in.

jtappa:
Updated.

README.md
```diff
@@ -17,95 +18,116 @@ This module adds a new exec provider capable of executing PowerShell commands.
...
+
+#####`logoutput`
+Defines whether to log command output in addition to logging the exit code. If you specify 'on_failure', it only logs the output when the command has an exit code that does not match any value specified by the `returns` attribute. Valid options: 'true', 'false', and 'on_failure'. Default: 'on_failure'.
+
+#####`onlyif`
+Runs the exec only if the command returns 0. Valid options: String. Default: Undefined.
+
+#####`path`
+Specifies the search path used for command execution. Valid options: String of the path, an array, or a semicolon-separated list. Default: Undefined. 
+
```
ferventcoder: :+1:

ferventcoder:
I'm +1 to merge. @joshcooper

joshcooper:
:+1:

PR #52, opened 2015-04-30
[DO NOT MERGE](maint) add AppVeyor

ferventcoder:
Add AppVeyor to PowerShell module.

- Matrix - Run against several supported versions of Puppet
- Notify the Ruby / Puppet versions
- Capture Environment vars / Gemfile.lock contents
- Ensure Administrator - AppVeyor doesn't have an Administrator account
by default, so put the adminstrator account back for specs to pass
- create top level Puppet data folders - puppet runs on Windows fails
until the folders exist

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -20,7 +20,7 @@ gem "facter", *location_for(ENV['FACTER_LOCATION'] || '>= 1')
 gem "hiera", *location_for(ENV['HIERA_LOCATION'] || '>= 1')
 
 beaker_version = ENV['BEAKER_VERSION']
-beaker_rspec_version = ENV['BEAKER_RSPEC_VERSION'] || '~> 4.0'
+beaker_rspec_version = ENV['BEAKER_RSPEC_VERSION'] #|| '~> 4.0'
 group :development, :test do
   gem 'rspec'
   gem 'mocha'
@@ -31,6 +31,8 @@ group :development, :test do
   gem 'simplecov',               :require => false
   if beaker_version
     gem 'beaker', *location_for(beaker_version)
+  else
+    gem 'beaker',                :require => false, :platforms => :ruby
   end
   if beaker_rspec_version
     gem 'beaker-rspec', *location_for(beaker_rspec_version)
@@ -43,6 +45,7 @@ end
 platforms :mswin, :mingw do
   gem "ffi", "~> 1.9", :require => false
   gem "sys-admin", "~> 1.5.6", :require => false
+  gem "win32-api", "~> 1.4.8", :require => false
   gem "win32-dir", "~> 0.3", :require => false
   gem "win32-eventlog", "~> 0.5", :require => false
   gem "win32-process", "~> 0.6", :require => false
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -36,14 +43,18 @@ end
...
+  gem "win32-api", "~> 1.4.8", :require => false
+  gem "win32-dir", "~> 0.3", :require => false
+  gem "win32-eventlog", "~> 0.5", :require => false
+  gem "win32-process", "~> 0.6", :require => false
+  gem "win32-security", "~> 0.1", :require => false
+  gem "win32-service", "~> 0.7", :require => false
+  gem "win32-taskscheduler", "~> 0.2", :require => false
+  gem "win32console", "~> 1.3", :require => false
+  gem "windows-api", "~> 0.4", :require => false
+  gem "windows-pr", "~> 1.2", :require => false
```
ferventcoder: Making this less restrictive so that bundler, using the puppet version, can figure out the exact version dependencies here.

appveyor.yml
```diff
@@ -0,0 +1,32 @@
...
+  - PUPPET_GEM_VERSION: 3.4.3
+    RUBY_VER: 193
+  - PUPPET_GEM_VERSION: 3.7.5
+    RUBY_VER: 193
+  - PUPPET_GEM_VERSION: 3.7.5
+    RUBY_VER: 200-x64
+  - PUPPET_GEM_VERSION: 4.0.0
+    RUBY_VER: 21
+  - PUPPET_GEM_VERSION: 4.0.0
+    RUBY_VER: 21-x64
```
ferventcoder: Test against the latest release of most minor versions, capturing the oldest, the newest and a release somewhere in between.

appveyor.yml
```diff
@@ -0,0 +1,32 @@
...
+install:
+- SET PATH=C:\Ruby%RUBY_VER%\bin;%PATH%
+- gem install bundler --quiet --no-ri --no-rdoc
+- bundle install --jobs 4 --retry 2 --without development
```
ferventcoder: This should really speed up bundle installs, with a retry if there are network issues.

appveyor.yml
```diff
@@ -0,0 +1,32 @@
...
+- type Gemfile.lock
+build: off
+test_script:
+- bundle exec puppet -V
+- ruby -v
+- bundle exec rake spec spec/unit -fd -b
```
ferventcoder: Format the spec output and provide a full backtrace.

appveyor.yml
```diff
@@ -0,0 +1,33 @@
...
+matrix:
+  fast_finish: true
+install:
+- SET PATH=C:\Ruby%RUBY_VER%\bin;%PATH%
+- gem update --system 2.3.0
+- gem install bundler --quiet --no-ri --no-rdoc
+- bundle install --jobs 4 --retry 2 --without development
+- type Gemfile.lock
+build: off
+test_script:
```
Iristyle: I'm not sure if we're standardizing our AppVeyor integrations, but seems like maybe we should be calling rake tasks?

ferventcoder: You'll love this one. `bundle exec rake spec` fails because beaker-rspec isn't installed.

ferventcoder: But also options available to pass to rspec are not immediately available to pass to a rake task. All the rake tasks are set up differently across the repositories. I'd see standardizing all of that before moving to a standard call. 

Right now it's much easier to stick without something we can control that isn't invasive.

ferventcoder:
Also, any failures here are real. This thing doesn't actually pass specs when run against windows.

ferventcoder:
Okay, we can't yet merge this - will need to update the commits.

ferventcoder:
Superseded by #60

PR #60, opened 2015-07-10
(MODULES-2207) Update Modulesync

ferventcoder:
This ensures the following items are under moduleysnc:

* .gitignore
* .travis.yml
* CONTRIBUTING.md
* Gemfile
* appveyor.yml
* spec/spec.opts

Most importantly, this ensures that acl has up to date Gemfile, Travis CI and
AppVeyor configurations that allow it to be tested properly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,25 +1,17 @@
-## MAC OS
-.DS_Store
-
-## TEXTMATE
-*.tmproj
-tmtags
-
-## EMACS
-*~
-\#*
-.\#*
-
-## VIM
-*.swp
-tags
-
-.rspec
-.*.sw[op]
-.bundle/
+pkg/
 Gemfile.lock
 Gemfile.local
-
-## IntelliJ / Rubymine
-.idea
+vendor/
+spec/fixtures/
+log/
+junit/
+.vagrant/
+.bundle/
+coverage/
+.idea/
+.metadata
 *.iml
+.*.sw[op]
+.DS_Store
+.rspec
+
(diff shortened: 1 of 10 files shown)


ferventcoder:
Expected failure on this one.

ferventcoder:
Well that was no fun to fix.

ferventcoder:
w00t! \0/ 

Finally...

ferventcoder:
@cyberious @Iristyle @joshcooper @jpogran this is ready to look at.

Iristyle:
So on 7d13076, we change our testing matrix to `3.4.0` for AppVeyor and TravisCI, but our metadata claims `3.x` compatibility at https://github.com/puppetlabs/puppetlabs-powershell/blob/master/metadata.json#L32

So are we concerned about a testing blind spot for older versions of Puppet?

ferventcoder:
@Iristyle thought we mentioned this elsewhere, but handling the metadata.json updates will be done with a separate ticket (for all repos affected). And IMHO the oldest version blind spot is low risk as compared to not having testing at all. We may decide to move the oldest version to 3.4.0 in the metadata.json updates ticket.

cyberious:
@ferventcoder considering this was not in puppetlabs namespace until 3.4 I would see no reason why 3.4.0 wouldn't be the oldest, also I believe the risk to be minimal between 3.0 and 3.4

cyberious:
:+1:

ferventcoder:
Pushed latest changes. Looks like I was able to back it all the way up to 3.0.0, which is the actual oldest supported.

ferventcoder:
have a bit of a tweak I'll be pushing - once that comes in and goes green I will notify.

cyberious:
gtg

PR #61, opened 2015-07-16
(MODULES-2207) Add Beaker-Rspec

ferventcoder:
Beaker-rspec is used by acceptance testing and should be present for acceptance
testing to properly pass.

### DIFF
--- a/.sync.yml
+++ b/.sync.yml
@@ -10,7 +10,7 @@ Gemfile:
     ':development':
       - gem: rake
       - gem: rspec
-        version: '~>2.14.0'
+        version: '~>2.14.1'
       - gem: puppet-lint
       - gem: puppetlabs_spec_helper
       - gem: puppet_facts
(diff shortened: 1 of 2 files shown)


ferventcoder:
@cyberious I haven't investigated why beaker-rspec suddenly causes issues with builds when included..

cyberious:
@ferventcoder not sure this is an issue with us, as much if you pin to 2.14.0 it didn't have any_instance_recorder_for which rspec-puppet uses, however 2.14.1 does have this, so I suggest we fix it to ~>2.14.1

ferventcoder:
@cyberious Pin what? We are not installing beaker-rspec b/c `--without system_tests` I'm guessing you are talking about rspec itself, because that is the only thing I see from the logs that installed 2.14.0.

ferventcoder:
Rspec has been moved up to ~> 2.14.1 and it appears that has fixed the issue as you stated @cyberious . Thanks!

PR #73, opened 2016-01-24
Mention that templates need to have DOS linefeeds

mattock:
Not knowing/remembering this bit me badly (see ticket [MODULES-3007](https://tickets.puppetlabs.com/browse/MODULES-3007)). I hope this small addition to README.md helps other people to save time and debugging effort.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ $obj = $(Get-WMIObject Win32_UserAccount -Filter "Name='Guest'")
 $obj.Rename("OtherGuest")
 ~~~
 
-This has the added benefit of not requiring escaping '$' in the PowerShell code.
+This has the added benefit of not requiring escaping '$' in the PowerShell code. Note that templates need to have DOS linefeeds or they will not work as expected. One tool for converting UNIX linefeeds to DOS linefeeds is [unix2dos](http://freecode.com/projects/dos2unix).
 
 ##Reference
 


README.md
```diff
@@ -72,7 +72,7 @@ $obj = $(Get-WMIObject Win32_UserAccount -Filter "Name='Guest'")
 $obj.Rename("OtherGuest")
 ~~~
 
-This has the added benefit of not requiring escaping '$' in the PowerShell code.
+This has the added benefit of not requiring escaping '$' in the PowerShell code. Note that templates need to have DOS linefeeds or they will not work as expected. One tool for converting UNIX linefeeds to DOS linefeeds is [unix2dos](http://freecode.com/projects/dos2unix).
```
ferventcoder: @joshcooper @Iristyle this is interesting to me - I wonder if this is required based on our current method of calling Posh.

mattock: The provider currently uses the "-Command - < "<template-file>" style call. It might be worth a try to test -File "<template-file>" instead and see if this problem is reproducible then.

ferventcoder: `-File` has issues with exit codes - specifically why we avoid it like the plague. We support POSH v2 and forward. If you want some reading to understand the reasoning - http://joshua.poehls.me/2012/powershell-script-module-boilerplate/

mattock: Yeah, I was not really suggesting that. The really nasty part in this is that no warning/error is printed if the template has linefeeds of the wrong kind. This makes debugging these problems really hard until one has experienced this once the hard way.

Regarding calling Posh... is it necessary to use cmd.exe as the middleman between Puppet and Powershell?

joshcooper: We should just fix the provider. If `content` contains the sequence `\r\n`, then open the tempfile in `binary` mode, preserving the existing line endings. Otherwise, `content` doesn't contain any line endings or is using `\n`, and we should open the tempfile in `text` mode, ensuring the resulting script has `\r\n`.

ferventcoder: Agree on fixing the provider.

PR #74, opened 2016-01-28
(MODULES-2962) Reuse PowerShell Session

jpogran:
This PR is a port of the PowerShell session code from the DSC module. This code includes some significant speed and system utilization improvements.

In the prior version every call to exec started a brand new PowerShell process for each of the exec modes: command, onlyif and unless. This was costly in both time and system resources as PowerShell takes significant time to intialize and frequent process creations spike cpu on the target system.

In this version, a single PowerShell process is created that lives not only for all the modes of exec, but also for the entirety of the Puppet run. We still have a one time cost of time and system utilization for
process creation for the entire run, but gain the ability to run consecutive commands.

A second improvement to this version is that we directly pass commands to PowerShell through shell redirection, and not through reading a file in like the last version. This avoids any of the parsing errors
(MODULES-2634) we have seen. This also improves syntax checking, as we now validate the passed in commands using PowerShell itself before the command is run, allowing us to catch errors sooner and report them.

Backwards compatibility was an important requirement for this PR as well. The same behavior the old way of executing PowerShell must be maintained with this new way. This presented a problem as the long-lived session would keep all variables from Puppet run to Puppet run. Care was taken to reset the environment between runs by resetting both the PowerShell variables and Environment variables to original state each run. This also refreshed the PowerShell and Environment variables, allowing one to set global variables and have them take effect in the next run, just like the old behavior could. 

While care was taken to run on all environments the old version did, we cannot use the new method with ruby 1.9.3. Ruby in this version used win32console, which breaks when stdout is redirected. This breaks the Puppet logging stream and halts the Puppet run. The code will detect if running on ruby 1.9.3 and use the old PowerShell invocation code, which while slower still allows the user to use this module.

This PR also removes support for Windows 2003, as Puppet itself has ceased support for Windows 2003.

### DIFF
--- a/lib/puppet/provider/exec/powershell.rb
+++ b/lib/puppet/provider/exec/powershell.rb
@@ -1,4 +1,7 @@
 require 'puppet/provider/exec'
+if Puppet::Util::Platform.windows? &&  Puppet.version > '3.8.5'
+  require_relative '../../../puppet_x/puppetlabs/powershell_manager'
+end
 
 Puppet::Type.type(:exec).provide :powershell, :parent => Puppet::Provider::Exec do
   confine :operatingsystem => :windows
@@ -25,17 +28,69 @@ Puppet::Type.type(:exec).provide :powershell, :parent => Puppet::Provider::Exec 
         }
   EOT
 
+  def self.upgrade_message
+    Puppet.warning <<-UPGRADE
+The current Puppet version is outdated and uses a library that was
+previously necessary on the current Ruby verison to support a colored console.
+
+Unfortunately this library prevents the PowerShell module from using a shared
+PowerShell process to dramatically improve the performance of resource
+application.
+
+To enable these improvements, it is suggested to upgrade to any x64 version of
+Puppet (including 3.x), or to a Puppet version newer than 3.x.
+    UPGRADE
+  end
+
+  def self.win32console_enabled?
+    @enabled ||= defined?(Win32::Console) && Win32::Console.class == Class
+  end
+
+  if self.win32console_enabled?
+    upgrade_message
+  end
+
+  def self.powershell_args
+    ps_args = ['-NoProfile', '-NonInteractive', '-Sta', '-NoLogo', '-ExecutionPolicy', 'Bypass', '-Command']
+    ps_args << '-' if !self.win32console_enabled?
+    ps_args
+  end
+
+  def ps_manager
+    PuppetX::PowerShell::PowerShellManager.instance("#{command(:powershell)} #{self.class.powershell_args.join(' ')}")
+  end
+
   def run(command, check = false)
-    write_script(command) do |native_path|
-      # Ideally, we could keep a handle open on the temp file in this
-      # process (to prevent TOCTOU attacks), and execute powershell
-      # with -File <path>. But powershell complains that it can't open
-      # the file for exclusive access. If we close the handle, then an
-      # attacker could modify the file before we invoke powershell. So
-      # we redirect powershell's stdin to read from the file. Current
-      # versions of Windows use per-user temp directories with strong
-      # permissions, but I'd rather not make (poor) assumptions.
-      return super("cmd.exe /c \"\"#{native_path(command(:powershell))}\" #{args} -Command - < \"#{native_path}\"\"", check)
+    if self.class.win32console_enabled?
+      write_script(command) do |native_path|
+        # Ideally, we could keep a handle open on the temp file in this
+        # process (to prevent TOCTOU attacks), and execute powershell
+        # with -File <path>. But powershell complains that it can't open
+        # the file for exclusive access. If we close the handle, then an
+        # attacker could modify the file before we invoke powershell. So
+        # we redirect powershell's stdin to read from the file. Current
+        # versions of Windows use per-user temp directories with strong
+        # permissions, but I'd rather not make (poor) assumptions.
+        return super("cmd.exe /c \"\"#{native_path(command(:powershell))}\" #{legacy_args} -Command - < \"#{native_path}\"\"", check)
+      end
+    else
+      result = ps_manager.execute(command)
+
+      stdout      = result[:stdout]
+      stderr       = result[:stderr]
+      exit_code = result[:exitcode]
+
+      unless stderr.nil?
+        stderr.each do |er|
+          er.each { |e| Puppet.debug "STDERR: #{e.chop}" } unless er.empty?
+        end
+      end
+
+      Puppet.debug "STDERR: #{result[:errormessage]}" unless result[:errormessage].nil?
+
+      output = Puppet::Util::Execution::ProcessOutput.new(stdout.to_s || '', exit_code)
+
+      return output, output
     end
   end
 
@@ -63,7 +118,7 @@ Puppet::Type.type(:exec).provide :powershell, :parent => Puppet::Provider::Exec 
     end
   end
 
-  def args
+  def legacy_args
     '-NoProfile -NonInteractive -NoLogo -ExecutionPolicy Bypass'
   end
 end
--- a/lib/puppet_x/puppetlabs/powershell_manager.rb
+++ b/lib/puppet_x/puppetlabs/powershell_manager.rb
@@ -1,0 +1,82 @@
+require 'spec_helper'
+require 'puppet/type'
+require 'puppet_x/puppetlabs/powershell_manager' if Puppet::Util::Platform.windows?
+
+module PuppetX
+  module Dsc
+    class PowerShellManager; end
+  end
+end
+
+describe PuppetX::Dsc::PowerShellManager,
+  :if => Puppet::Util::Platform.windows? && !Facter.value(:uses_win32console) do
+
+  let (:manager) {
+    powershell = Puppet::Type.type(:base_dsc).defaultprovider.command(:powershell)
+    powershell_args = Puppet::Type.type(:base_dsc).defaultprovider.powershell_args
+    PuppetX::Dsc::PowerShellManager.instance("#{powershell} #{powershell_args.join(' ')}")
+  }
+
+  describe "when provided powershell commands" do
+    it "should return simple output" do
+      result = manager.execute('write-output foo')[:stdout]
+      expect(result).to eq("foo\n")
+    end
+
+    it "should execute cmdlets" do
+      result = manager.execute('ls')[:stdout]
+      expect(result).not_to eq(nil)
+    end
+
+    it "should execute cmdlets with pipes" do
+      result = manager.execute('Get-Process | ? { $_.PID -ne $PID }')[:stdout]
+      expect(result).not_to eq(nil)
+    end
+
+    it "should execute multi-line" do
+      result = manager.execute(<<-CODE
+$foo = ls
+$count = $foo.count
+$count
+      CODE
+      )[:stdout]
+      expect(result).not_to eq(nil)
+    end
+
+    it "should reuse the same PowerShell process for multiple calls" do
+      first_pid = manager.execute('[Diagnostics.Process]::GetCurrentProcess().Id')[:stdout]
+      second_pid = manager.execute('[Diagnostics.Process]::GetCurrentProcess().Id')[:stdout]
+
+      expect(first_pid).to eq(second_pid)
+    end
+
+    it "should be able to write more than the 64k default buffer size to child process stdout without deadlocking the Ruby parent process" do
+      result = manager.execute(<<-CODE
+$bytes_in_k = (1024 * 64) + 1
+[Text.Encoding]::UTF8.GetString((New-Object Byte[] ($bytes_in_k))) | Write-Output
+        CODE
+        )[:stdout]
+      expect(result).not_to eq(nil)
+    end
+
+    it "should return a JSON response with a timeout error if the execution timeout is exceeded" do
+      timeout_ms = 100
+      result = manager.execute('sleep 1', timeout_ms)
+      response = JSON.parse(result[:stdout])
+      msg = /Catastrophic failure\: PowerShell DSC resource timeout \(#{timeout_ms} ms\) exceeded while executing/
+      expect(response['errormessage']).to match(msg)
+    end
+
+    it "should not deadlock and return a valid JSON response given invalid unparseable PowerShell code" do
+      result = manager.execute(<<-CODE
+        {
+
+        CODE
+        )
+
+      response = JSON.parse(result[:stdout])
+      expect(response["errormessage"]).not_to be_empty
+    end
+  end
+
+end
--- a/lib/puppet_x/puppetlabs/powershell_manager.rb
+++ b/lib/puppet_x/puppetlabs/powershell_manager.rb
@@ -1,82 +1,208 @@
-require 'spec_helper'
-require 'puppet/type'
-require 'puppet_x/puppetlabs/powershell_manager' if Puppet::Util::Platform.windows?
+require 'rexml/document'
+require 'securerandom'
+require 'open3'
+require 'base64'
+require 'ffi'
 
 module PuppetX
-  module Dsc
-    class PowerShellManager; end
-  end
-end
+  module PowerShell
+    class PowerShellManager
+      extend Puppet::Util::Windows::String
+      extend FFI::Library
 
-describe PuppetX::Dsc::PowerShellManager,
-  :if => Puppet::Util::Platform.windows? && !Facter.value(:uses_win32console) do
+      @@instances = {}
 
-  let (:manager) {
-    powershell = Puppet::Type.type(:base_dsc).defaultprovider.command(:powershell)
-    powershell_args = Puppet::Type.type(:base_dsc).defaultprovider.powershell_args
-    PuppetX::Dsc::PowerShellManager.instance("#{powershell} #{powershell_args.join(' ')}")
-  }
+      def self.instance(cmd)
+        @@instances[:cmd] ||= PowerShellManager.new(cmd)
+      end
 
-  describe "when provided powershell commands" do
-    it "should return simple output" do
-      result = manager.execute('write-output foo')[:stdout]
-      expect(result).to eq("foo\n")
-    end
+      def initialize(cmd)
+        @stdin, @stdout, @stderr, @ps_process = Open3.popen3(cmd)
 
-    it "should execute cmdlets" do
-      result = manager.execute('ls')[:stdout]
-      expect(result).not_to eq(nil)
-    end
+        Puppet.debug "#{Time.now} #{cmd} is running as pid: #{@ps_process[:pid]}"
 
-    it "should execute cmdlets with pipes" do
-      result = manager.execute('Get-Process | ? { $_.PID -ne $PID }')[:stdout]
-      expect(result).not_to eq(nil)
-    end
+        at_exit { exit }
+      end
 
-    it "should execute multi-line" do
-      result = manager.execute(<<-CODE
-$foo = ls
-$count = $foo.count
-$count
-      CODE
-      )[:stdout]
-      expect(result).not_to eq(nil)
-    end
+      def execute(powershell_code, timeout_ms = 300 * 1000)
+        output_ready_event_name =  "Global\\#{SecureRandom.uuid}"
+        output_ready_event = self.class.create_event(output_ready_event_name)
 
-    it "should reuse the same PowerShell process for multiple calls" do
-      first_pid = manager.execute('[Diagnostics.Process]::GetCurrentProcess().Id')[:stdout]
-      second_pid = manager.execute('[Diagnostics.Process]::GetCurrentProcess().Id')[:stdout]
+        code = make_ps_code(powershell_code, output_ready_event_name, timeout_ms)
 
-      expect(first_pid).to eq(second_pid)
-    end
+        out, err = exec_read_result(code, output_ready_event)
 
-    it "should be able to write more than the 64k default buffer size to child process stdout without deadlocking the Ruby parent process" do
-      result = manager.execute(<<-CODE
-$bytes_in_k = (1024 * 64) + 1
-[Text.Encoding]::UTF8.GetString((New-Object Byte[] ($bytes_in_k))) | Write-Output
-        CODE
-        )[:stdout]
-      expect(result).not_to eq(nil)
-    end
+        # Powershell adds in newline characters as it tries to wrap output around the display (by default 80 chars).
+        # This behavior is expected and cannot be changed, however it corrupts the XML e.g. newlines in the middle of
+        # element names; So instead, part of the XML is Base64 encoded prior to being put on STDOUT and in ruby all
+        # newline characters are stripped. Then where required decoded from Base64 back into text
+        out = REXML::Document.new(out.gsub(/\n/,""))
 
-    it "should return a JSON response with a timeout error if the execution timeout is exceeded" do
-      timeout_ms = 100
-      result = manager.execute('sleep 1', timeout_ms)
-      response = JSON.parse(result[:stdout])
-      msg = /Catastrophic failure\: PowerShell DSC resource timeout \(#{timeout_ms} ms\) exceeded while executing/
-      expect(response['errormessage']).to match(msg)
-    end
+        hash = {}
+        REXML::XPath.each(out, "//Property").map { |x| [x.attributes["Name"], x.text] }.each do | prop |
+          hash[prop[0]] = prop[1]
+        end
 
-    it "should not deadlock and return a valid JSON response given invalid unparseable PowerShell code" do
-      result = manager.execute(<<-CODE
         {
+          :exitcode     => hash['exitcode'].to_i,
+          :errormessage => hash['errormessage'].nil? ? nil : Base64.decode64(hash['errormessage']),
+          :stdout       => hash['stdout'].nil? ? nil : Base64.decode64(hash['stdout']),
+          :stderr       => err,
+        }
+      ensure
+        FFI::WIN32.CloseHandle(output_ready_event) if output_ready_event
+      end
 
-        CODE
-        )
+      def exit
+        Puppet.debug "PowerShellManager exiting..."
+        @stdin.puts "\nexit\n"
+        @stdin.close
+        @stdout.close
+        @stderr.close
 
-      response = JSON.parse(result[:stdout])
-      expect(response["errormessage"]).not_to be_empty
+        exit_msg = "PowerShell process did not terminate in reasonable time"
+        begin
+          Timeout.timeout(3) do
+            Puppet.debug "Awaiting PowerShell process termination..."
+            @exit_status = @ps_process.value
+          end
+        rescue Timeout::Error
+        end
+
+        exit_msg = "PowerShell process exited: #{@exit_status}" if @exit_status
+        Puppet.debug(exit_msg)
+        if @ps_process.alive?
+          Puppet.debug("Forcefully terminating PowerShell process.")
+          Process.kill('KILL', @ps_process[:pid])
+        end
+      end
+
+      def template_path
+        File.expand_path('../../templates', __FILE__)
+      end
+
+      def make_ps_code(powershell_code, output_ready_event_name, timeout_ms = 300 * 1000)
+        template_file = File.new(template_path + "/invoke_ps_command.erb").read
+        template = ERB.new(template_file, nil, '-')
+        template.result(binding)
+      end
+
+      private
+
+      def self.is_readable?(stream, timeout = 0.5)
+        read_ready = IO.select([stream], [], [], timeout)
+        read_ready && stream == read_ready[0][0]
+      end
+
+      def self.create_event(name, manual_reset = false, initial_state = false)
+        handle = FFI::Pointer::NULL_HANDLE
+
+        FFI::Pointer.from_string_to_wide_string(name) do |name_ptr|
+          handle = CreateEventW(FFI::Pointer::NULL,
+            manual_reset ? 1 : FFI::WIN32_FALSE,
+            initial_state ? 1 : FFI::WIN32_FALSE,
+            name_ptr)
+
+          if handle == FFI::Pointer::NULL_HANDLE
+            msg = "Failed to create new event #{name}"
+            raise Puppet::Util::Windows::Error.new(msg)
+          end
+        end
+
+        handle
+      end
+
+      WAIT_ABANDONED = 0x00000080
+      WAIT_OBJECT_0 = 0x00000000
+      WAIT_TIMEOUT = 0x00000102
+      WAIT_FAILED = 0xFFFFFFFF
+
+      def self.wait_on(wait_object, timeout_ms = 50)
+        wait_result = Puppet::Util::Windows::Process::WaitForSingleObject(
+          wait_object, timeout_ms)
+
+        case wait_result
+        when WAIT_OBJECT_0
+          Puppet.debug "Wait object signaled"
+        when WAIT_TIMEOUT
+          Puppet.debug "Waited #{timeout_ms} milliseconds..."
+        # only applicable to mutexes - should never happen here
+        when WAIT_ABANDONED
+          msg = 'Catastrophic failure: wait object in inconsistent state'
+          raise Puppet::Util::Windows::Error.new(msg)
+        when WAIT_FAILED
+          msg = 'Catastrophic failure: waiting on object to be signaled'
+          raise Puppet::Util::Windows::Error.new(msg)
+        end
+
+        wait_result
+      end
+
+      def write_stdin(input)
+        @stdin.puts(input)
+      rescue => e
+        msg = "Error writing STDIN / reading STDOUT: #{e}"
+        raise Puppet::Util::Windows::Error.new(msg)
+      end
+
+      def drain_pipe(pipe, iterations = 10)
+        output = []
+        0.upto(10) do
+          break if !self.class.is_readable?(pipe, 0.1)
+          l = pipe.gets
+          Puppet.debug "#{Time.now} PIPE> #{l}"
+          output << l
+        end
+        output
+      end
+
+      def read_stdout(output_ready_event, wait_interval_ms = 50)
+        output = []
+        errors = []
+        waited = 0
+
+        # drain the pipe while waiting for the event signal
+        while WAIT_TIMEOUT == self.class.wait_on(output_ready_event, wait_interval_ms)
+          # TODO: While this does ensure that both pipes have been
+          # drained it can block on either longer than necessary or
+          # deadlock waiting for one or the other to finish. The correct
+          # way to deal with this is to drain each pipe from seperate threads
+          # but time ran on in this implementation and this will be addressed soon
+          output << drain_pipe(@stdout)
+          errors << drain_pipe(@stderr)
+          waited += wait_interval_ms
+        end
+
+        Puppet.debug "Waited #{waited} total milliseconds."
+
+        # once signaled, ensure everything has been drained
+        output << drain_pipe(@stdout, 1000)
+        errors << drain_pipe(@stderr, 1000)
+
+        errors = errors.reject { |e| e.empty? }
+
+        return output.join(''), errors
+      rescue => e
+        msg = "Error reading PIPE: #{e}"
+        raise Puppet::Util::Windows::Error.new(msg)
+      end
+
+      def exec_read_result(powershell_code, output_ready_event)
+        write_stdin(powershell_code)
+        read_stdout(output_ready_event)
+      end
+
+      ffi_convention :stdcall
+
+      # https://msdn.microsoft.com/en-us/library/windows/desktop/ms682396(v=vs.85).aspx
+      # HANDLE WINAPI CreateEvent(
+      #   _In_opt_ LPSECURITY_ATTRIBUTES lpEventAttributes,
+      #   _In_     BOOL                  bManualReset,
+      #   _In_     BOOL                  bInitialState,
+      #   _In_opt_ LPCTSTR               lpName
+      # );
+      ffi_lib :kernel32
+      attach_function_private :CreateEventW, [:pointer, :win32_bool, :win32_bool, :lpcwstr], :handle
     end
   end
-
 end
--- a/lib/puppet_x/templates/invoke_ps_command.erb
+++ b/lib/puppet_x/templates/invoke_ps_command.erb
@@ -1,0 +1,259 @@
+$hostSource = @"
+using System;
+using System.Globalization;
+using System.Management.Automation.Host;
+using System.Threading;
+
+public class PuppetPSHost : PSHost
+{
+    private Guid _hostId = Guid.NewGuid();
+    private bool shouldExit;
+    private int exitCode;
+
+    public PuppetPSHost () {}
+
+    public bool ShouldExit
+    {
+        get { return this.shouldExit; }
+    }
+
+    public int ExitCode
+    {
+        get { return this.exitCode; }
+    }
+
+    public void ResetExitStatus()
+    {
+      this.exitCode = 0;
+      this.shouldExit = false;
+    }
+
+    public override Guid InstanceId
+    {
+        get { return _hostId; }
+    }
+
+    public override string Name
+    {
+        get { return "PuppetPSHost"; }
+    }
+
+    public override Version Version
+    {
+        get { return new Version(1, 0); }
+    }
+
+    public override PSHostUserInterface UI
+    {
+        get { return null; }
+    }
+
+    public override CultureInfo CurrentCulture
+    {
+        get { return Thread.CurrentThread.CurrentCulture; }
+    }
+
+    public override CultureInfo CurrentUICulture
+    {
+        get { return Thread.CurrentThread.CurrentUICulture; }
+    }
+
+    public override void EnterNestedPrompt()
+    {
+        throw new NotImplementedException();
+    }
+
+    public override void ExitNestedPrompt()
+    {
+        throw new NotImplementedException();
+    }
+
+    public override void NotifyBeginApplication()
+    {
+        return;
+    }
+
+    public override void NotifyEndApplication()
+    {
+        return;
+    }
+
+    public override void SetShouldExit(int exitCode)
+    {
+      this.shouldExit = true;
+      this.exitCode = exitCode;
+    }
+}
+"@
+
+function New-XmlResult
+{
+  param(
+    [Parameter()]$exitcode,
+    [Parameter()]$output,
+    [Parameter()]$errormessage
+  )
+
+  # we make our own xml because ConvertTo-Xml makes hard to parse xml ruby side
+  # and we need to be sure
+  $xml = [xml]@"
+<ReturnResult>
+  <Property Name='exitcode'>$($exitcode)</Property>
+  <Property Name='errormessage'>$([System.Convert]::ToBase64String([System.Text.Encoding]::UTF8.GetBytes([string]$errormessage)))</Property>
+  <Property Name='stdout'>$([System.Convert]::ToBase64String([System.Text.Encoding]::UTF8.GetBytes([string]$output)))</Property>
+</ReturnResult>
+"@
+  $xml.OuterXml
+}
+
+#this is a string so we can import into our dynamic PS instance
+$ourFunctions = @'
+function Get-ProcessEnvironmentVariables
+{
+  $processVars = [Environment]::GetEnvironmentVariables('Process').Keys |
+    % -Begin { $h = @{} } -Process { $h.$_ = (Get-Item Env:\$_).Value } -End { $h }
+
+  # eliminate Machine / User vars so that we have only process vars
+  'Machine', 'User' |
+    % { [Environment]::GetEnvironmentVariables($_).GetEnumerator() } |
+    ? { $processVars.ContainsKey($_.Name) -and ($processVars[$_.Name] -eq $_.Value) } |
+    % { $processVars.Remove($_.Name) }
+
+  $processVars.GetEnumerator() | Sort-Object Name
+}
+
+function Reset-ProcessEnvironmentVariables
+{
+  param($processVars)
+
+  # query Machine vars from registry, ensuring expansion EXCEPT for PATH
+  $vars = [Environment]::GetEnvironmentVariables('Machine').GetEnumerator() |
+    % -Begin { $h = @{} } -Process { $v = if ($_.Name -eq 'Path') { $_.Value } else { [Environment]::GetEnvironmentVariable($_.Name, 'Machine') }; $h."$($_.Name)" = $v } -End { $h }
+
+  # query User vars from registry, ensuring expansion EXCEPT for PATH
+  [Environment]::GetEnvironmentVariables('User').GetEnumerator() | % {
+      if ($_.Name -eq 'Path') { $vars[$_.Name] += ';' + $_.Value }
+      else
+      {
+        $value = [Environment]::GetEnvironmentVariable($_.Name, 'User')
+        $vars[$_.Name] = $value
+      }
+    }
+
+  $processVars.GetEnumerator() | % { $vars[$_.Name] = $_.Value }
+
+  Remove-Item -Path Env:\* -ErrorAction SilentlyContinue -WarningAction SilentlyContinue -Recurse -Verbose
+
+  $vars.GetEnumerator() | % { Set-Item -Path "Env:\$($_.Name)" -Value $_.Value -Verbose }
+}
+
+function Reset-ProcessPowerShellVariables
+{
+  param($psVariables)
+  $psVariables | %{
+    $tempVar = $_
+    if(-not(Get-Variable -Name $_.Name -ErrorAction SilentlyContinue)){
+      New-Variable -Name $_.Name -Value $_.Value -Description $_.Description -Option $_.Options -Visibility $_.Visibility
+    }
+  }
+}
+'@
+
+Add-Type -TypeDefinition $hostSource -Language CSharp
+
+$event = [System.Threading.EventWaitHandle]::OpenExisting("<%= output_ready_event_name %>")
+
+if ($runspace -eq $null){
+  # CreateDefault2 requires PS3
+  if ([System.Management.Automation.Runspaces.InitialSessionState].GetMethod('CreateDefault2')){
+    $sessionState = [System.Management.Automation.Runspaces.InitialSessionState]::CreateDefault2()
+  }else{
+    $sessionState = [System.Management.Automation.Runspaces.InitialSessionState]::CreateDefault()
+  }
+
+  $puppetPSHost = New-Object PuppetPSHost
+  $runspace = [System.Management.Automation.Runspaces.RunspaceFactory]::CreateRunspace($puppetPSHost, $sessionState)
+  $runspace.Open()
+}
+
+$powershell_code = @'
+<%= powershell_code %>
+'@
+
+try
+{
+  $ps = $null
+  $puppetPSHost.ResetExitStatus()
+
+  if($PSVersionTable.PSVersion -ge [Version]'3.0'){
+    $runspace.ResetRunspaceState()
+  }
+
+  $ps = [System.Management.Automation.PowerShell]::Create()
+  $ps.Runspace = $runspace
+
+  [Void]$ps.AddScript($ourFunctions)
+  $ps.Invoke()
+
+  if(!$environmentVariables){
+    $environmentVariables = $ps.AddCommand('Get-ProcessEnvironmentVariables').Invoke()
+  }
+
+  if($PSVersionTable.PSVersion -le [Version]'2.0'){
+    if(!$psVariables){
+      $psVariables = $ps.AddScript('Get-Variable').Invoke()
+    }
+    [void]$ps.AddScript('Get-Variable -Scope Global | Remove-Variable -Force -ErrorAction SilentlyContinue -WarningAction SilentlyContinue')
+    $ps.Invoke()
+    [void]$ps.AddCommand('Reset-ProcessPowerShellVariables').AddParameter('psVariables', $psVariables)
+    $ps.Invoke()
+  }
+
+  [Void]$ps.AddCommand('Reset-ProcessEnvironmentVariables').AddParameter('processVars', $environmentVariables)
+  $ps.Invoke()
+
+  [Void]$ps.AddScript($powershell_code)
+  $asyncResult = $ps.BeginInvoke()
+
+  if (!$asyncResult.AsyncWaitHandle.WaitOne(<%= timeout_ms %>, $false)){
+    throw "Catastrophic failure: PowerShell module timeout (<%= timeout_ms %> ms) exceeded while executing"
+  }
+
+  $output = $ps.EndInvoke($asyncResult)
+  $output = $output | Out-String
+
+  New-XmlResult -exitcode $puppetPSHost.Exitcode -output $output -errormessage $null
+}
+catch
+{
+  try
+  {
+    if ($runspace) { $runspace.Dispose() }
+  }
+  finally
+  {
+    $runspace = $null
+  }
+  if(($puppetPSHost -ne $null) -and $puppetPSHost.ExitCode){
+    $ec = $puppetPSHost.ExitCode
+  }else{
+    # This is technically not true at this point as we do not
+    # know what exitcode we should return as an unexpected exception
+    # happened and the user did not set an exitcode. Our best guess
+    # is to return 1 so that we ensure Puppet reports this run as an error.
+    $ec = 1
+  }
+  $output = $_.Exception.Message | Out-String
+  New-XmlResult -exitcode $ec -output $null -errormessage $output
+}
+finally
+{
+  [Void]$event.Set()
+  [Void]$event.Close()
+  if ($PSVersionTable.CLRVersion.Major -ge 3) {
+    [Void]$event.Dispose()
+  }
+  if ($ps -ne $null) { [Void]$ps.Dispose() }
+}
+
+# always need a trailing newline to ensure PowerShell parses code
--- a/spec/integration/puppet_x/puppetlabs/powershell_manager_spec.rb
+++ b/spec/integration/puppet_x/puppetlabs/powershell_manager_spec.rb
@@ -1,0 +1,82 @@
+require 'spec_helper'
+require 'puppet/type'
+require 'puppet_x/puppetlabs/powershell_manager' if Puppet::Util::Platform.windows?
+
+module PuppetX
+  module Dsc
+    class PowerShellManager; end
+  end
+end
+
+describe PuppetX::Dsc::PowerShellManager,
+  :if => Puppet::Util::Platform.windows? && !Facter.value(:uses_win32console) do
+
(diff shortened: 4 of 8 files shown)


lib/puppet_x/puppetlabs/powershell_manager.rb
```diff
@@ -0,0 +1,244 @@
...
+          $asyncResult = $ps.BeginInvoke()
+
+          if (!$asyncResult.AsyncWaitHandle.WaitOne(#{timeout_ms}, $false))
+          {
+            throw "Catastrophic failure: PowerShell DSC resource timeout (#{timeout_ms} ms) exceeded while executing"
+          }
+
+          $output = $ps.EndInvoke($asyncResult)
+
+          if($ps.HadErrors){
```
jpogran: Catch non terminating errors. However, did we treat non-terminating as terminating before?

lib/puppet_x/puppetlabs/powershell_manager.rb
```diff
@@ -0,0 +1,244 @@
...
+      extend FFI::Library
+
+      @@instances = {}
+
+      def self.instance(cmd)
+        @@instances[:cmd] ||= PowerShellManager.new(cmd)
+      end
+
+      def initialize(cmd)
+        @stdin, @stdout, @ps_process = Open3.popen2(cmd)
```
jpogran: popen2 doesn't provide stderr

lib/puppet/provider/exec/powershell.rb
```diff
@@ -25,17 +26,55 @@
...
+        # the file for exclusive access. If we close the handle, then an
+        # attacker could modify the file before we invoke powershell. So
+        # we redirect powershell's stdin to read from the file. Current
+        # versions of Windows use per-user temp directories with strong
+        # permissions, but I'd rather not make (poor) assumptions.
+        return super("cmd.exe /c \"\"#{native_path(command(:powershell))}\" #{args} -Command - < \"#{native_path}\"\"", check)
+      end
+    else
+      output = ps_manager.execute(command)[:stdout]
+      output = JSON.parse(output)
```
jpogran: Do we want to do this? JSON output makes it easy to pass data bag, but turns the PS output to JSON as well which means we lose the orginal PS formatting. Is it really expected to 'return' output from a PS command?

lib/puppet_x/puppetlabs/powershell_manager.rb
```diff
@@ -0,0 +1,244 @@
...
+          {
+            if ($runspace) { $runspace.Dispose() }
+          }
+          finally
+          {
+            $runspace = $null
+          }
+          @{
+            output       = $null
+            exitcode     = 1  # TODO: this is lying
```
jpogran: We are technically lying here. Since we keep PS open, there isn't any exit code or return code. Only binaries return exit codes in PS, cmdlets and the like do not.

lib/puppet/provider/exec/powershell.rb
```diff
@@ -25,17 +26,55 @@
...
+      write_script(command) do |native_path|
+        # Ideally, we could keep a handle open on the temp file in this
+        # process (to prevent TOCTOU attacks), and execute powershell
+        # with -File <path>. But powershell complains that it can't open
+        # the file for exclusive access. If we close the handle, then an
+        # attacker could modify the file before we invoke powershell. So
+        # we redirect powershell's stdin to read from the file. Current
+        # versions of Windows use per-user temp directories with strong
+        # permissions, but I'd rather not make (poor) assumptions.
+        return super("cmd.exe /c \"\"#{native_path(command(:powershell))}\" #{args} -Command - < \"#{native_path}\"\"", check)
```
ferventcoder: I'm fine with - but I think we need to not write a file out somewhere first. We should be able to bind the file into memory and pass that to the call.

lib/puppet_x/puppetlabs/powershell_manager.rb
ferventcoder: Right, we need to be doing this in a more robust way, we need popen3.

lib/puppet_x/puppetlabs/powershell_manager.rb
ferventcoder: Non-terminating errors are not considered bad errors. There are too many things that will run and go up as non-terminating errors but are completely successful.

ferventcoder: More like non-terminating errors are warnings.

lib/puppet_x/puppetlabs/powershell_manager.rb
ferventcoder: Return the exit code from the command or function, if there is one.

ferventcoder: Not sure how easily possible this is.

lib/puppet_x/puppetlabs/powershell_manager.rb
jpogran: We decided offline not to report on non-terminating errors

lib/puppet_x/puppetlabs/powershell_manager.rb
jpogran: Stay tuned

lib/puppet/provider/exec/powershell.rb
Iristyle: Writing the file is still OK as we do it safely.  As it turns out, it also appears necessary to allow the user to continue to use `exit` in their scripts to generate an exit code.

To spell it out a little more, doing the following will exit the `Powershell` instance, and there's no way for us to capture the exit code at this point:

```powershell
$ps.AddScript('exit 1')
$output = $ps.Invoke()
```

However, given a file `foo.ps1`

```powershell
exit 1
```

And an execution like:

```powershell
$ps.AddScript('./foo.ps1')
$output = $ps.Invoke()

$ps.AddScript('echo $LASTEXITCODE')
$exitcode = $ps.Invoke()
```

We can actually get to the exit code.

lib/puppet_x/puppetlabs/powershell_version.rb
```diff
@@ -0,0 +1,49 @@
+module PuppetX
+  module PowerShell
+    class PowerShellVersion
+    end
+  end
+end
+
+if Puppet::Util::Platform.windows?
```
Iristyle: I thought we were yanking this?

lib/puppet/provider/exec/powershell.rb
```diff
@@ -25,17 +26,63 @@
...
+Unfortunately this library prevents the PowerShell module from using a shared
+PowerShell process to dramatically improve the performance of resource
+application.
+
+To enable these improvements, it is suggested to upgrade to any x64 version of
+Puppet (including 3.x), or to a Puppet version newer than 3.x.
+    UPGRADE
+  end
+
+  if self.win32console_enabled?
```
Iristyle: Used before being defined...

```
Undefined method `win32console_enabled?
```

lib/puppet_x/puppetlabs/powershell_manager.rb
```diff
@@ -0,0 +1,240 @@
...
+        end
+        output
+      end
+
+      def read_stdout(output_ready_event, wait_interval_ms = 50)
+        output = []
+        waited = 0
+
+        # drain the pipe while waiting for the event signal
+        while WAIT_TIMEOUT == self.class.wait_on(output_ready_event, wait_interval_ms)
```
Iristyle: We need to drain both `stdout` and `stderr` now...

lib/puppet_x/puppetlabs/powershell_manager.rb
```diff
@@ -0,0 +1,240 @@
...
+        @stdin, @stdout, @ps_process = Open3.popen2(cmd)
+
+        Puppet.debug "#{Time.now} #{cmd} is running as pid: #{@ps_process[:pid]}"
+
+        at_exit { exit }
+      end
+
+      def execute(powershell_code, timeout_ms = 300 * 1000)
+        output_ready_event_name =  "Global\\#{SecureRandom.uuid}"
+        output_ready_event = self.class.create_event(output_ready_event_name)
```
Iristyle: We need another event for `stderr`

lib/puppet_x/puppetlabs/powershell_version.rb
jpogran: Missed, removed

lib/puppet/provider/exec/powershell.rb
jpogran: One last thing to change, is that we couldn't use a file created by Ruby, as it kept PowerShell from being able to read the file. Creating and then reading inside PowerShell fixed this

lib/puppet_x/puppetlabs/powershell_manager.rb
jpogran: We have fixed this in all but this place. This is the only place that has a hardcoded exitcode now,

lib/puppet/provider/exec/powershell.rb
ferventcoder: :+1:

lib/puppet_x/puppetlabs/powershell_manager.rb
```diff
@@ -0,0 +1,247 @@
...
+
+        # always need a trailing newline to ensure PowerShell parses code
+        code = <<-CODE
```
jpogran: Is there a way we caan break this out to a separate file or method? This is getting long and could get problematic in future @Iristyle @ferventcoder @glennsarti

lib/puppet/provider/exec/powershell.rb
ferventcoder: For this particular bit, I think the piped redirection is the problem here. That is another problem entirely that is documented in https://tickets.puppetlabs.com/browse/MODULES-2634 and https://tickets.puppetlabs.com/browse/MODULES-2774

ferventcoder: AND not really the focus of this PR.

lib/puppet/provider/exec/powershell.rb
Iristyle: Output from PS is only for internal usage .. not really used by the module.  All we *really* care about is exit code.

lib/puppet_x/puppetlabs/powershell_manager.rb
```diff
@@ -0,0 +1,247 @@
...
+          $runspace = [RunspaceFactory]::CreateRunspace($sessionState)
+          $runspace.Open()
+        }
+
+        $powershell_code = @'
+#{powershell_code}
+'@
+        try
+        {
+          $file = [IO.Path]::GetTempFileName()
```
Iristyle: This method of writing files is less secure than it can be.  At the very least, we should use `[IO.Path]::GetTempPath()` and [[IO.Path]::GetRandomFileName](https://msdn.microsoft.com/en-us/library/system.io.path.getrandomfilename\(v=vs.110\).aspx)

I'm testing locally to try and find a way to keep the file open and have PowerShell read it, but not having much luck.  Also, I think there's some potential for running into encoding issues by passing the string into PowerShell and having it write the file, so I wonder if we can use some additional options when writing the file on the Ruby side to influence creation - the Ruby implementation lives roughly at https://github.com/ruby/ruby/blob/eee616315d406e91a232c33434f5195d72578b65/win32/win32.c#L6041-L6210

lib/puppet_x/puppetlabs/powershell_manager.rb
```diff
@@ -0,0 +1,268 @@
...
+          $ps = $null
+
+          $file = [IO.Path]::GetTempFileName()
+          $file = $file -replace ".tmp",".ps1"
+          $fileStream = [IO.File]::Open($file, [IO.FileMode]::OpenOrCreate, [IO.FileAccess]::Write, [IO.FileShare]::Read)
+          $sw = New-Object IO.StreamWriter ($fileStream, [Text.Encoding]::UTF8)
+          $sw.Write($powershell_code)
+          $sw.Flush()
+          $sw.Close()
+          $sw.Dispose()
```
Iristyle: So I took another pass at figuring out how to capture exit codes, and I was just able to finally determine that we **can** get at the exit code properly if we implement a custom `PSHost`.  If we do that, then the file writing becomes moot as it's no longer necessary to get at `$LASTEXITCODE` (though obviously there are some other changes that need to be made).

I wrote a real quick proof of concept demonstrating that:
https://gist.github.com/Iristyle/0a63ef1bd7840244a201

/cc @puppetlabs/windows

ferventcoder: I would have suggested that earlier but I didn't know what kind of undertaking you wanted to do here.

ferventcoder: Commented on your POC.

Rakefile
```diff
@@ -10,6 +10,13 @@ RSpec::Core::RakeTask.new(:unit) do |t,args|
   t.verbose     = true
 end
 
+desc "Integration tests"
```
Iristyle: There isn't any real precedent for segregating this AFAIK (unless you found an example somewhere?), so I'd prefer to just stay consistent with how this works in the other modules:

* https://github.com/puppetlabs/puppetlabs-acl/blob/master/Rakefile
* https://github.com/puppetlabs/puppetlabs-registry/blob/master/Rakefile
* https://github.com/puppetlabs/puppetlabs-reboot/blob/master/Rakefile
* https://github.com/puppetlabs/puppetlabs-sqlserver/blob/master/Rakefile
* https://github.com/puppetlabs/puppetlabs-wsus_client/blob/master/Rakefile

lib/puppet_x/templates/invoke_ps_command.erb
```diff
@@ -0,0 +1,165 @@
...
+  if (!$asyncResult.AsyncWaitHandle.WaitOne(<%= timeout_ms %>, $false))
+  {
+    throw "Catastrophic failure: PowerShell resource timeout (<%= timeout_ms %> ms) exceeded while executing"
+  }
+
+  # doesn't generate output - just exits
+  $output = $ps.EndInvoke($asyncResult)
+
+  return @{
+    output       = $output
```
Iristyle: Should we just rename this to `stdout`?

Also - are we doing anything with `$Error`?

lib/puppet/provider/exec/powershell.rb
```diff
@@ -25,18 +26,60 @@
...
-      # permissions, but I'd rather not make (poor) assumptions.
-      return super("cmd.exe /c \"\"#{native_path(command(:powershell))}\" #{args} -Command - < \"#{native_path}\"\"", check)
+    result = ps_manager.execute(command)
+    stdout = result[:stdout]
+    stderr = result[:stderr]
+
+    output = JSON.parse(stdout)
+
+    exit_status = output['exitcode']
+    if exit_status.nil?
```
Iristyle: Our current code doesn't allow a nil `exitcode`

lib/puppet/provider/exec/powershell.rb
```diff
@@ -25,18 +26,60 @@
...
+
+    unless stderr.nil?
+      stderr.each do |er|
+        er.each { |e| Puppet.debug "STDERR: #{e.chop}" } unless er.empty?
+      end
     end
+
+    # TODO: figure out how to handle json text here if at all
+    stuff = Puppet::Util::Execution::ProcessOutput.new(output['output'].to_s || '', exit_status)
+    return stuff, stuff
```
Iristyle: Sweet... `stuff, stuff`!

lib/puppet/provider/exec/powershell.rb
```diff
@@ -25,18 +26,60 @@
...
-      # process (to prevent TOCTOU attacks), and execute powershell
-      # with -File <path>. But powershell complains that it can't open
-      # the file for exclusive access. If we close the handle, then an
-      # attacker could modify the file before we invoke powershell. So
-      # we redirect powershell's stdin to read from the file. Current
-      # versions of Windows use per-user temp directories with strong
-      # permissions, but I'd rather not make (poor) assumptions.
-      return super("cmd.exe /c \"\"#{native_path(command(:powershell))}\" #{args} -Command - < \"#{native_path}\"\"", check)
+    result = ps_manager.execute(command)
+    stdout = result[:stdout]
```
Iristyle: This is not really `stdout`... but I think that we *should* make it so that is.  Another note below about restructuring the return value from `PowerShellManager.execute`

lib/puppet_x/puppetlabs/powershell_manager.rb
```diff
@@ -0,0 +1,181 @@
...
+
+      def execute(powershell_code, timeout_ms = 300 * 1000)
+        output_ready_event_name =  "Global\\#{SecureRandom.uuid}"
+        output_ready_event = self.class.create_event(output_ready_event_name)
+
+        code = make_ps_code(powershell_code, output_ready_event_name, timeout_ms)
+
+        out, err = exec_read_result(code, output_ready_event)
+
+        { :stdout => out, :stderr => err }
```
Iristyle: I think that we should reshape the object here to make a bit more sense, like this:

```ruby
out = JSON.parse(out)

{
  :exitcode => out['exitcode'],
  :errormessage => out['errormessage'],
  :stdout => out['output'],
  :stderr => err
}
```

Then the downstream consumer doesn't need to do any JSON parsing, and the members are really indicative of what they are.

spec/integration/puppet_x/puppetlabs/powershell_manager_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+    end
+
+    it "should be able to write more than the 64k default buffer size to child process stdout without deadlocking the Ruby parent process" do
+      result = manager.execute(<<-CODE
+$bytes_in_k = (1024 * 64) + 1
+[Text.Encoding]::UTF8.GetString((New-Object Byte[] ($bytes_in_k))) | Write-Output
+        CODE
+        )[:stdout]
+      expect(result).not_to eq(nil)
+    end
```
Iristyle: We also need to do a test that pumps to `stdout` and `stderr` at the same time... and ensures we don't deadlock.

lib/puppet/provider/exec/powershell.rb
jpogran: Barring the silly name, following the lead from https://github.com/puppetlabs/puppet/blob/master/lib/puppet/provider/exec.rb#L74-L77

lib/puppet/provider/exec/powershell.rb
jpogran: Correct, now it doesn't. Removing

lib/puppet/provider/exec/powershell.rb
jpogran: Agreed. Changing

lib/puppet_x/puppetlabs/powershell_manager.rb
jpogran: Agreed. Changing

lib/puppet_x/templates/invoke_ps_command.erb
jpogran: Renamed to stdout.

I didn't do anything with `$Error` because its a hash of all the errors in the current session, with no differentiation between non-terminating and terminating errors. If it's a terminating error, it would be caught by our catch block and reported there. If it was non-terminating, we decided to continue not doing anything with them. So I didn't see the need for doing anything with `$Error`. Am I missing something?

lib/puppet/provider/exec/powershell.rb
```diff
@@ -25,18 +26,60 @@
...
+
+    exit_status = output['exitcode']
+    if exit_status.nil?
+      # TODO: Should we warn user they did not return an exitcode?
+      exit_status = 0
+    end
+
+    unless stderr.nil?
+      stderr.each do |er|
+        er.each { |e| Puppet.debug "STDERR: #{e.chop}" } unless er.empty?
```
jpogran: Any problems with this block handling stderr output?

Rakefile
jpogran: Is there a reason @glennsarti ?

ferventcoder:
Fully qualify all types used. I'm testing this with POSH v2 and it fails on `InitialSessionState`.

~~~puppet
exec { 'Write Posh Location':
  command   => 'get-process powershell* | %{ Write-Host $($_.Path)}',
  provider  => powershell,
  logoutput => true
}
~~~

~~~sh
=> default: Error: /Stage[main]/Main/Exec[Write Posh Location]/returns: change from notrun to 0 failed: A JSON text must at least contain two octets!
==> default: Debug: Waited 50 milliseconds...
==> default: Debug: 2016-02-11 20:15:26 +0000 PIPE> Unable to find type [InitialSessionState]: make sure that the assembly containi
==> default:
==> default:
==> default: Debug: 2016-02-11 20:15:26 +0000 PIPE> ng this type is loaded.
==> default:
==> default: Debug: 2016-02-11 20:15:26 +0000 PIPE> At line:4 char:28
==> default:
==> default: Debug: 2016-02-11 20:15:26 +0000 PIPE> +   if ([InitialSessionState] <<<< .GetMethod('CreateDefault2'))
==> default:
==> default: Debug: 2016-02-11 20:15:26 +0000 PIPE>     + CategoryInfo          : InvalidOperation: (InitialSessionState:String) [
==> default:
==> default: Debug: 2016-02-11 20:15:26 +0000 PIPE>    ], RuntimeException
==> default:
==> default: Debug: 2016-02-11 20:15:26 +0000 PIPE>     + FullyQualifiedErrorId : TypeNotFound
==> default:
==> default: Debug: 2016-02-11 20:15:26 +0000 PIPE>
==> default:
==> default: Debug: 2016-02-11 20:15:26 +0000 PIPE> Method invocation failed because [PuppetPSHost] doesn't contain a method named
==> default:
==> default: Debug: 2016-02-11 20:15:26 +0000 PIPE> 'new'.
==> default:
==> default: Debug: 2016-02-11 20:15:26 +0000 PIPE> At line:12 char:38
==> default:
==> default: Debug: 2016-02-11 20:15:26 +0000 PIPE> +   $puppetPSHost = [PuppetPSHost]::new <<<< ()
==> default:
==> default: Debug: 2016-02-11 20:15:26 +0000 PIPE>     + CategoryInfo          : InvalidOperation: (new:String) [], RuntimeExcept
==> default:
==> default: Debug: 2016-02-11 20:15:26 +0000 PIPE>    ion
==> default:
==> default: Debug: 2016-02-11 20:15:26 +0000 PIPE>     + FullyQualifiedErrorId : MethodNotFound
==> default:
==> default: Debug: 2016-02-11 20:15:26 +0000 PIPE>
==> default:
==> default: Debug: 2016-02-11 20:15:26 +0000 PIPE> Multiple ambiguous overloads found for "CreateRunspace" and the argument count:
==> default:
==> default: Debug: 2016-02-11 20:15:26 +0000 PIPE>  "2".
==> default:
==> default: Debug: 2016-02-11 20:15:26 +0000 PIPE> At line:13 char:48
==> default:
==> default: Debug: 2016-02-11 20:15:26 +0000 PIPE> +   $runspace = [RunspaceFactory]::CreateRunspace <<<< ($puppetPSHost, $session
==> default:
==> default: Debug: 2016-02-11 20:15:26 +0000 PIPE> State)
==> default:
==> default: Debug: 2016-02-11 20:15:26 +0000 PIPE>     + CategoryInfo          : NotSpecified: (:) [], MethodException
==> default:
==> default: Debug: 2016-02-11 20:15:26 +0000 PIPE>     + FullyQualifiedErrorId : MethodCountCouldNotFindBest
==> default:
==> default: Debug: 2016-02-11 20:15:26 +0000 PIPE>
==> default:
==> default: Debug: 2016-02-11 20:15:26 +0000 PIPE> You cannot call a method on a null-valued expression.
==> default:
==> default: Debug: 2016-02-11 20:15:26 +0000 PIPE> At line:14 char:17
==> default:
==> default: Debug: 2016-02-11 20:15:26 +0000 PIPE> +   $runspace.Open <<<< ()
==> default:
==> default: Debug: 2016-02-11 20:15:26 +0000 PIPE>     + CategoryInfo          : InvalidOperation: (Open:String) [], RuntimeExcep
==> default:
==> default: Debug: 2016-02-11 20:15:26 +0000 PIPE>    tion
==> default:
==> default: Debug: 2016-02-11 20:15:26 +0000 PIPE>     + FullyQualifiedErrorId : InvokeMethodOnNull
==> default:
==> default: Debug: 2016-02-11 20:15:26 +0000 PIPE>
==> default:
==> default: Debug: 2016-02-11 20:15:26 +0000 PIPE> Method invocation failed because [System.Threading.EventWaitHandle] doesn't con
==> default:
==> default: Debug: 2016-02-11 20:15:26 +0000 PIPE> tain a method named 'Dispose'.
==> default:
==> default: Debug: 2016-02-11 20:15:26 +0000 PIPE> At line:46 char:23
==> default:
==> default: Debug: 2016-02-11 20:15:26 +0000 PIPE> +   [Void]$event.Dispose <<<< ()
==> default:
==> default: Debug: 2016-02-11 20:15:26 +0000 PIPE>     + CategoryInfo          : InvalidOperation: (Dispose:String) [], RuntimeEx
==> default:
==> default: Debug: 2016-02-11 20:15:26 +0000 PIPE>    ception
==> default:
==> default: Debug: 2016-02-11 20:15:26 +0000 PIPE>     + FullyQualifiedErrorId : MethodNotFound
==> default:
==> default: Debug: 2016-02-11 20:15:26 +0000 PIPE>
==> default:
==> default: Debug: 2016-02-11 20:15:26 +0000 PIPE> The term 'ConvertTo-Json' is not recognized as the name of a cmdlet, function,
==> default:
==> default: Debug: 2016-02-11 20:15:26 +0000 PIPE> script file, or operable program. Check the spelling of the name, or if a path
==> default:
==> default: Debug: 2016-02-11 20:15:26 +0000 PIPE> was included, verify that the path is correct and try again.
==> default:
==> default: Debug: 2016-02-11 20:15:26 +0000 PIPE> At line:41 char:21
==> default:
==> default: Debug: 2016-02-11 20:15:26 +0000 PIPE> +   } | ConvertTo-Json <<<<  -Compress
==> default:
==> default: Debug: 2016-02-11 20:15:26 +0000 PIPE>     + CategoryInfo          : ObjectNotFound: (ConvertTo-Json:String) [], Comm
==> default:
==> default: Debug: 2016-02-11 20:15:26 +0000 PIPE>    andNotFoundException
==> default:
==> default: Debug: 2016-02-11 20:15:26 +0000 PIPE>     + FullyQualifiedErrorId : CommandNotFoundException
==> default:
==> default: Debug: 2016-02-11 20:15:26 +0000 PIPE>
==> default:
==> default: Debug: Wait object signaled
==> default:
==> default: Debug: Waited 50 total milliseconds.
==> default: Error: A JSON text must at least contain two octets!
==> default:
==> default: c:/Program Files/Puppet Labs/Puppet/sys/ruby/lib/ruby/2.0.0/json/common.rb:155:in `initialize'
==> default: c:/Program Files/Puppet Labs/Puppet/sys/ruby/lib/ruby/2.0.0/json/common.rb:155:in `new'
==> default: c:/Program Files/Puppet Labs/Puppet/sys/ruby/lib/ruby/2.0.0/json/common.rb:155:in `parse'
==> default: C:/tmp/vagrant-puppet/modules-b7052f0cdc4f82c87f2a22e65b072c98/powershell/lib/puppet_x/puppetlabs/powershell_manager.rb:33:in `execute'
==> default: C:/tmp/vagrant-puppet/modules-b7052f0cdc4f82c87f2a22e65b072c98/powershell/lib/puppet/provider/exec/powershell.rb:62:in `run'
==> default: c:/Program Files/Puppet Labs/Puppet/puppet/lib/puppet/type/exec.rb:130:in `block in sync'
==> default: c:/Program Files/Puppet Labs/Puppet/puppet/lib/puppet/type/exec.rb:127:in `times'
==> default: c:/Program Files/Puppet Labs/Puppet/puppet/lib/puppet/type/exec.rb:127:in `sync'
==> default: c:/Program Files/Puppet Labs/Puppet/puppet/lib/puppet/transaction/resource_harness.rb:204:in `sync'
==> default: c:/Program Files/Puppet Labs/Puppet/puppet/lib/puppet/transaction/resource_harness.rb:128:in `sync_if_needed'
==> default: c:/Program Files/Puppet Labs/Puppet/puppet/lib/puppet/transaction/resource_harness.rb:89:in `block in perform_changes'
==> default: c:/Program Files/Puppet Labs/Puppet/puppet/lib/puppet/transaction/resource_harness.rb:88:in `each'
==> default: c:/Program Files/Puppet Labs/Puppet/puppet/lib/puppet/transaction/resource_harness.rb:88:in `perform_changes'
==> default: c:/Program Files/Puppet Labs/Puppet/puppet/lib/puppet/transaction/resource_harness.rb:20:in `evaluate'
==> default: c:/Program Files/Puppet Labs/Puppet/puppet/lib/puppet/transaction.rb:204:in `apply'
==> default: c:/Program Files/Puppet Labs/Puppet/puppet/lib/puppet/transaction.rb:217:in `eval_resource'
==> default: c:/Program Files/Puppet Labs/Puppet/puppet/lib/puppet/transaction.rb:147:in `call'
==> default: c:/Program Files/Puppet Labs/Puppet/puppet/lib/puppet/transaction.rb:147:in `block (2 levels) in evaluate'
==> default: c:/Program Files/Puppet Labs/Puppet/puppet/lib/puppet/util.rb:335:in `block in thinmark'
==> default: c:/Program Files/Puppet Labs/Puppet/sys/ruby/lib/ruby/2.0.0/benchmark.rb:296:in `realtime'
==> default: c:/Program Files/Puppet Labs/Puppet/puppet/lib/puppet/util.rb:334:in `thinmark'
==> default: c:/Program Files/Puppet Labs/Puppet/puppet/lib/puppet/transaction.rb:147:in `block in evaluate'
==> default: c:/Program Files/Puppet Labs/Puppet/puppet/lib/puppet/graph/relationship_graph.rb:118:in `traverse'
==> default: c:/Program Files/Puppet Labs/Puppet/puppet/lib/puppet/transaction.rb:138:in `evaluate'
==> default: c:/Program Files/Puppet Labs/Puppet/puppet/lib/puppet/resource/catalog.rb:169:in `block in apply'
==> default: c:/Program Files/Puppet Labs/Puppet/puppet/lib/puppet/util/log.rb:149:in `with_destination'
==> default: c:/Program Files/Puppet Labs/Puppet/puppet/lib/puppet/transaction/report.rb:112:in `as_logging_destination'
==> default: c:/Program Files/Puppet Labs/Puppet/puppet/lib/puppet/resource/catalog.rb:168:in `apply'
==> default: c:/Program Files/Puppet Labs/Puppet/puppet/lib/puppet/configurer.rb:120:in `block in apply_catalog'
==> default: c:/Program Files/Puppet Labs/Puppet/puppet/lib/puppet/util.rb:161:in `block in benchmark'
==> default: c:/Program Files/Puppet Labs/Puppet/sys/ruby/lib/ruby/2.0.0/benchmark.rb:296:in `realtime'
==> default: c:/Program Files/Puppet Labs/Puppet/puppet/lib/puppet/util.rb:160:in `benchmark'
==> default: c:/Program Files/Puppet Labs/Puppet/puppet/lib/puppet/configurer.rb:119:in `apply_catalog'
==> default: c:/Program Files/Puppet Labs/Puppet/puppet/lib/puppet/configurer.rb:227:in `run_internal'
==> default: c:/Program Files/Puppet Labs/Puppet/puppet/lib/puppet/configurer.rb:134:in `block in run'
==> default: c:/Program Files/Puppet Labs/Puppet/puppet/lib/puppet/context.rb:64:in `override'
==> default: c:/Program Files/Puppet Labs/Puppet/puppet/lib/puppet.rb:246:in `override'
==> default: c:/Program Files/Puppet Labs/Puppet/puppet/lib/puppet/configurer.rb:133:in `run'
==> default: c:/Program Files/Puppet Labs/Puppet/puppet/lib/puppet/application/apply.rb:302:in `apply_catalog'
==> default: c:/Program Files/Puppet Labs/Puppet/puppet/lib/puppet/application/apply.rb:236:in `block in main'
==> default: c:/Program Files/Puppet Labs/Puppet/puppet/lib/puppet/context.rb:64:in `override'
==> default: c:/Program Files/Puppet Labs/Puppet/puppet/lib/puppet.rb:246:in `override'
==> default: c:/Program Files/Puppet Labs/Puppet/puppet/lib/puppet/application/apply.rb:198:in `main'
==> default: c:/Program Files/Puppet Labs/Puppet/puppet/lib/puppet/application/apply.rb:159:in `run_command'
==> default: c:/Program Files/Puppet Labs/Puppet/puppet/lib/puppet/application.rb:381:in `block (2 levels) in run'
==> default: c:/Program Files/Puppet Labs/Puppet/puppet/lib/puppet/application.rb:507:in `plugin_hook'
==> default: c:/Program Files/Puppet Labs/Puppet/puppet/lib/puppet/application.rb:381:in `block in run'
==> default: c:/Program Files/Puppet Labs/Puppet/puppet/lib/puppet/util.rb:496:in `exit_on_fail'
==> default: c:/Program Files/Puppet Labs/Puppet/puppet/lib/puppet/application.rb:381:in `run'
==> default: c:/Program Files/Puppet Labs/Puppet/puppet/lib/puppet/util/command_line.rb:146:in `run'
==> default: c:/Program Files/Puppet Labs/Puppet/puppet/lib/puppet/util/command_line.rb:92:in `execute'
==> default: c:/Program Files/Puppet Labs/Puppet/puppet/bin/puppet:8:in `<main>'
==> default: Error: /Stage[main]/Main/Exec[Write Path]/returns: change from notrun to 0 failed: A JSON text must at least contain two octets!
==> default:
~~~

lib/puppet_x/templates/invoke_ps_command.erb
```diff
@@ -0,0 +1,165 @@
...
+    {
+      this.shouldExit = true;
+      this.exitCode = exitCode;
+    }
+}
+"@
+
+Add-Type -TypeDefinition $hostSource -Language CSharp
+
+$event = [Threading.EventWaitHandle]::OpenExisting("<%= output_ready_event_name %>")
```
ferventcoder: Fully qualify this.

lib/puppet_x/templates/invoke_ps_command.erb
```diff
@@ -0,0 +1,165 @@
...
+
+if ($runspace -eq $null)
+{
+  # CreateDefault2 requires PS3
+  if ([InitialSessionState].GetMethod('CreateDefault2'))
```
ferventcoder: Needs Fully qualified for POSH v2
`System.Management.Automation`?

Rakefile
glennsarti: I ported the tests from the DSC module which were placed in the integration directory.  There was no existing rake task for it, so I created one.

lib/puppet_x/templates/invoke_ps_command.erb
```diff
@@ -0,0 +1,179 @@
...
+  }
+
+  $output = $ps.EndInvoke($asyncResult)
+
+@"
+{
+  "stdout": "",
+  "exitcode": $($puppetPSHost.ExitCode),
+  "errormessage": ""
+}
```
jpogran: We have to make our own JSON here, as ConvertTo-JSON exists in PS v3 and above, not in v2. This is WIP

ferventcoder: We had a possibility of http://stackoverflow.com/a/29689642/18475, but it requires .NET 3.5 to be installed.

lib/puppet_x/templates/invoke_ps_command.erb
```diff
@@ -0,0 +1,225 @@
...
+  $xRoot.AppendChild($xElement) | Out-Null
+
+  $memStream = New-Object System.IO.MemoryStream
+  $writerSettings = New-Object System.Xml.XmlWriterSettings
+  $writerSettings.Indent = $true
+  $writerSettings.Encoding = New-Object System.Text.UTF8Encoding -Arg $false
+  $writerSettings.ConformanceLevel = [System.Xml.ConformanceLevel]::Document
+
+  $writer = [System.Xml.XmlWriter]::Create($memStream, $writerSettings)
+  $xDoc.Save($writer)
```
jpogran: This is still throwing an exception for the output from our buffer overlfow test. Despite having UTF8 specified everywhere I can think of

lib/puppet_x/templates/invoke_ps_command.erb
```diff
@@ -0,0 +1,226 @@
...
+
+  [System.Text.Encoding]::UTF8.GetString($memStream.ToArray())
+}
+
+$cleanupfunction = @"
+function Remove-ScriptVariables($path) {
```
Iristyle: Is there an advantage to clearing like this ... vs:

* Capture variable space at startup and stash somewhere (i.e. in a random GUID named variable)
* Restore runspace variables to capture prior to each execution

I just want to make sure I understand the implementation choice.

jpogran: In doing research I've seen your suggested approach (store before and compare to after) and my approach (search and find variables). I'm trying to have as few Invoke()'s as possible and avoid throwing exceptions.

Store Before/CompareAfter
pro: Remove-Variable can exclude an array of names, so easy oneliners
cons: At least two seperate Invoke()'s are needed outside of the primary Invoke()

Search and Destroy
pro: uses PowerShell's own parser to find variables and remove (silent if cant). One Invoke()
cons: relies on clever code

Iristyle: Don't forget that removing extraneous isn't good enough.  We also need to reset existing vars back to their original state where necessary.

lib/puppet/provider/exec/powershell.rb
```diff
@@ -25,17 +28,67 @@
...
-      # Ideally, we could keep a handle open on the temp file in this
-      # process (to prevent TOCTOU attacks), and execute powershell
-      # with -File <path>. But powershell complains that it can't open
-      # the file for exclusive access. If we close the handle, then an
-      # attacker could modify the file before we invoke powershell. So
-      # we redirect powershell's stdin to read from the file. Current
-      # versions of Windows use per-user temp directories with strong
-      # permissions, but I'd rather not make (poor) assumptions.
-      return super("cmd.exe /c \"\"#{native_path(command(:powershell))}\" #{args} -Command - < \"#{native_path}\"\"", check)
+    if defined?(Win32::Console) && Win32::Console.class == Class
```
jpogran: I need some help on this line. n00b ruby mistake I'm sure. I cannot use win32console_enabled? here, says method undefined. @ferventcoder @Iristyle

Iristyle: `self.class.win32console_enabled?` is what you're looking for I believe

lib/puppet/provider/exec/powershell.rb
```diff
@@ -25,17 +28,69 @@
...
+        stderr.each do |er|
+          er.each { |e| Puppet.debug "STDERR: #{e.chop}" } unless er.empty?
+        end
+      end
+
+      Puppet.debug "STDERR: #{result[:errormessage]}" unless result[:errormessage].nil?
+
+      stuff = Puppet::Util::Execution::ProcessOutput.new(stdout.to_s || '', exit_code)
+
+      return stuff, stuff
```
Iristyle: Umm...

lib/puppet_x/puppetlabs/powershell_manager.rb
```diff
@@ -0,0 +1,203 @@
...
+
+      def read_stdout(output_ready_event, wait_interval_ms = 50)
+        output = []
+        errors = []
+        waited = 0
+
+        # drain the pipe while waiting for the event signal
+        while WAIT_TIMEOUT == self.class.wait_on(output_ready_event, wait_interval_ms)
+          output << drain_pipe(@stdout)
+          errors << drain_pipe(@stderr)
```
Iristyle: I'm not sure that this is sufficient to prevent a deadlock - I *think* that these would actually need to alternate and in some way limit the amount of times that the `is_readable?` check is performed.

lib/puppet_x/puppetlabs/powershell_manager.rb
```diff
@@ -0,0 +1,203 @@
...
+        code = make_ps_code(powershell_code, output_ready_event_name, timeout_ms)
+
+        out, err = exec_read_result(code, output_ready_event)
+
+        # Powershell adds in newline characters as it tries to wrap output around the display (by default 80 chars).
+        # This behavior is expected and cannot be changed, however it corrupts the XML e.g. newlines in the middle of
+        # element names; So instead, part of the XML is Base64 encoded prior to being put on STDOUT and in ruby all
+        # newline characters are stripped. Then where required decoded from Base64 back into text
+        out = REXML::Document.new(out.gsub(/\n/,""))
+        #out = REXML::XPath.each(out, "//Property").map { |x| [x.attributes["Name"], x.text] }.to_h
```
Iristyle: Comment?

lib/puppet_x/puppetlabs/powershell_manager.rb
```diff
@@ -0,0 +1,203 @@
...
+        hash = {}
+        REXML::XPath.each(out, "//Property").map { |x| [x.attributes["Name"], x.text] }.each do | prop |
+          hash[prop[0]] = prop[1]
+        end
+
+        #require "pry";binding.pry
```
Iristyle: Comment?

lib/puppet_x/templates/invoke_ps_command.erb
```diff
@@ -0,0 +1,284 @@
...
+    if ($runspace) { $runspace.Dispose() }
+  }
+  finally
+  {
+    $runspace = $null
+  }
+  if(($puppetPSHost -ne $null) -and $puppetPSHost.ExitCode){
+      $ec = $puppetPSHost.ExitCode
+  }else{
+      $ec = 1  # TODO: this is lying
```
Iristyle: TODO - ticket or comment.. or both?

lib/puppet/provider/exec/powershell.rb
jpogran: Which part is 'umm...'? The poor naming or the array return? The array return comes from mimicking the exec provider https://github.com/puppetlabs/puppet/blob/master/lib/puppet/provider/exec.rb#L74-L77

lib/puppet_x/templates/invoke_ps_command.erb
```diff
@@ -0,0 +1,284 @@
...
+{
+  param(
+    [Parameter()]$exitcode,
+    [Parameter()]$output,
+    [Parameter()]$errormessage
+  )
+
+  # we make our own xml because ConvertTo-Xml makes hard to parse xml ruby side
+  # and we need to be sure
+  [System.Xml.XmlDocument]$xDoc = New-Object System.Xml.XmlDocument
```
Iristyle: So I think I understand the reasoning behind writing the stream in a specific manner back to Ruby (except for the indent), but I can't say that I understand building the structure like this... but maybe there was a reason? (I can't figure why all the casting though...)

This should work just fine and achieve the same result, if we really need to leave out children that are null:

```powershell
$xml = [xml]@"
<ReturnResult>
  <Property Name='exitCode'>$exitCode</Property>
</ReturnResult>
"@

if ($errormessage -ne $null)
{
  $prop = $xml.ReturnResult.FirstChild.Clone()
  $prop.Name = 'errormessage'
  $prop.InnerText = [Convert]::ToBase64String([Text.Encoding]::UTF8.GetBytes($errormessage))
  $xml.ReturnResult.AppendChild($prop) | Out-Null
}

if ($output -ne $null)
{
  $prop = $xml.ReturnResult.FirstChild.Clone()
  $prop.Name = 'stdout'
  $prop.InnerText = [Convert]::ToBase64String([Text.Encoding]::UTF8.GetBytes($output))
  $xml.ReturnResult.AppendChild($prop) | Out-Null
}
```

Even that feels a lot clunkier than necessary, and could really be something as simple as this, which is a lot easier to understand IMHO:

```powershell
$xml = [xml]@"
<ReturnResult>
  <Property Name='exitCode'>$exitCode</Property>
  <Property Name='errormessage'>$([Convert]::ToBase64String([Text.Encoding]::UTF8.GetBytes([string]$errormessage)))</Property>
  <Property Name='stdout'>$([Convert]::ToBase64String([Text.Encoding]::UTF8.GetBytes([string]$output)))</Property>
</ReturnResult>
"@
```

We can trivially change the Ruby receiving side here to deal with an empty string instead of a missing node for `errormessage` and `stdout`

lib/puppet/provider/exec/powershell.rb
Iristyle: The variable naming ;0

lib/puppet_x/templates/invoke_ps_command.erb
```diff
@@ -0,0 +1,284 @@
...
+    $output = [System.Convert]::ToBase64String([System.Text.Encoding]::UTF8.GetBytes($output))
+    [System.Xml.XmlElement]$xElement = $xDoc.CreateElement("Property")
+    $xElement.SetAttribute('Name','stdout')
+    $xElement.innerText = $output
+    $xRoot.AppendChild($xElement) | Out-Null
+  }
+
+  $memStream = New-Object System.IO.MemoryStream
+  $writerSettings = New-Object System.Xml.XmlWriterSettings
+  $writerSettings.Indent = $true
```
Iristyle: I don't think this ever shows up in the Ruby side of things, so don't think it's necessary

lib/puppet_x/templates/invoke_ps_command.erb
```diff
@@ -0,0 +1,284 @@
...
+  $writerSettings.Encoding = New-Object System.Text.UTF8Encoding -Arg $false
+  $writerSettings.ConformanceLevel = [System.Xml.ConformanceLevel]::Document
```
Iristyle: So `ConformanceLevel` can be `Document`, `Fragment` or `Auto` ... and the default is `Document` per https://msdn.microsoft.com/en-us/library/system.xml.xmlwritersettings.conformancelevel(v=vs.110).aspx.  Was there a reason we need to specify this explicitly?

lib/puppet_x/templates/invoke_ps_command.erb
```diff
@@ -0,0 +1,284 @@
...
+    [System.Xml.XmlElement]$xElement = $xDoc.CreateElement("Property")
+    $xElement.SetAttribute('Name','stdout')
+    $xElement.innerText = $output
+    $xRoot.AppendChild($xElement) | Out-Null
+  }
+
+  $memStream = New-Object System.IO.MemoryStream
+  $writerSettings = New-Object System.Xml.XmlWriterSettings
+  $writerSettings.Indent = $true
+  $writerSettings.Encoding = New-Object System.Text.UTF8Encoding -Arg $false
```
Iristyle: UTF8 is the default per https://msdn.microsoft.com/en-us/library/system.xml.xmlwritersettings.encoding(v=vs.110).aspx

> The text encoding to use. The default is Encoding.UTF8

lib/puppet_x/templates/invoke_ps_command.erb
```diff
@@ -0,0 +1,284 @@
...
+  $writerSettings.ConformanceLevel = [System.Xml.ConformanceLevel]::Document
+
+  $writer = [System.Xml.XmlWriter]::Create($memStream, $writerSettings)
```
Iristyle: It doesn't look to me like we actually need to use `$writerSettings` at all... maybe a leftover from before the Base64 encoding?

lib/puppet_x/templates/invoke_ps_command.erb
```diff
@@ -0,0 +1,284 @@
...
+  $xDoc.Save($writer)
+
+  [System.Text.Encoding]::UTF8.GetString($memStream.ToArray())
```
Iristyle: I think in the end `$xml.OuterXml` would give you the same result....

lib/puppet_x/templates/invoke_ps_command.erb
```diff
@@ -0,0 +1,284 @@
...
+}
+
+#this is a string so we can import into our dynamic PS instance
+$ourFunctions = @'
+function Get-EnvironmentVariables
+{
+  $processVars = [Environment]::GetEnvironmentVariables('Process').Keys |
+    % -Begin { $h = @{} } -Process { $h.$_ = (Get-Item Env:\$_).Value } -End { $h }
+
+  # Write-Host "Initially captured process vars:`n`n$($processVars | Out-String)"
```
Iristyle: Commented code can be axed

lib/puppet_x/templates/invoke_ps_command.erb
```diff
@@ -0,0 +1,284 @@
...
+  $ps = [System.Management.Automation.PowerShell]::Create()
+  $ps.Runspace = $runspace
+
+  [Void]$ps.AddScript($ourFunctions)
+  $ps.Invoke()
+
+  if(!$environmentVariables){
+    $environmentVariables = $ps.AddCommand('Get-EnvironmentVariables').Invoke()
+  }
+  if(!$psVariables){
```
Iristyle: I think we only need to capture these on PS <= 2

lib/puppet_x/templates/invoke_ps_command.erb
```diff
@@ -0,0 +1,284 @@
...
+
+  $writer = [System.Xml.XmlWriter]::Create($memStream, $writerSettings)
+  $xDoc.Save($writer)
+
+  [System.Text.Encoding]::UTF8.GetString($memStream.ToArray())
+}
+
+#this is a string so we can import into our dynamic PS instance
+$ourFunctions = @'
+function Get-EnvironmentVariables
```
Iristyle: I would rename this to `Get-ProcessEnvironmentVariables`, since just the name would lead me to believe it's all ENV vars.

lib/puppet_x/templates/invoke_ps_command.erb
jpogran: That was the original code, until we ran into problems with non XML escaped text. Then we went the CDATA route, using tags. This didn't escape everything, Glenn suggested the CDATA methods on the XDocument object, so built everything by hand. XDocument outputted utf-16 XML regardless of the docs saying default is utf8 (http://www.timvw.be/2007/01/08/generating-utf-8-with-systemxmlxmlwriter/), thus all the setup and XMLWriter stuff. That still didn't help with all possible characters, but most things. Newlines kept creeping in inside the XML tags, causing XML parsing issues, thus the attempts at indenting to avoid wrapping. Wrapping turned out to be from PSHost not XMLWriter. After abandoning CDATA for Base64, all issues were resolved, but XML left as is when issues with different Ruby and Puppet versions surfaced.

lib/puppet_x/templates/invoke_ps_command.erb
jpogran: Yes, part of an effort to make it output utf8 XML that was CDATA encoded.

lib/puppet_x/templates/invoke_ps_command.erb
jpogran: Yes, it was part of an effort to make it output utf8 XML that was CDATA encoded and wouldn't wrap.

lib/puppet_x/templates/invoke_ps_command.erb
jpogran: Incorrect. See many attempts by others http://www.timvw.be/2007/01/08/generating-utf-8-with-systemxmlxmlwriter

lib/puppet_x/templates/invoke_ps_command.erb
jpogran: Yes, leftover from before Base64, but was needed to ensure we actually got utf8 text. http://www.timvw.be/2007/01/08/generating-utf-8-with-systemxmlxmlwriter/

lib/puppet_x/templates/invoke_ps_command.erb
jpogran: No, it wouldn't, as the PSHost would wrap it

lib/puppet_x/templates/invoke_ps_command.erb
jpogran: Agreed

lib/puppet_x/puppetlabs/powershell_manager.rb
```diff
@@ -0,0 +1,202 @@
...
+      def write_stdin(input)
+        @stdin.puts(input)
+      rescue => e
+        msg = "Error writing STDIN / reading STDOUT: #{e}"
+        raise Puppet::Util::Windows::Error.new(msg)
+      end
+
+      def drain_pipe(pipe)
+        output = []
+        while self.class.is_readable?(pipe, 0.1) do
```
glennsarti: As per @Iristyle 's comment in read_stdout, I'm not sure if we want to go down the alternating path as that seems like we'd need two threads reading the pipes.  Perhaps the simpler answer is just to ensure that drain_pipe has a bounded loop.  e.g. A simple countdown where it will only loop in the while for, say, 100 times before exiting.  As it stands, this loop is unbound and, in theory, could just spin in this loop indefinitely (as long as there's data in STDOUT)

Iristyle: @glennsarti you are correct that we actually need 2 separate threads draining `stdout` and `stderr` to implement this properly.  Our hokey timeslicing-esque approach is better than nothing, but it's wrong.

This was one of the nice things in DSC about only having to read `stdout` - we were able to dodge some of that complexity.

I think we bound the iterations for now, but flesh out the threaded solution in a subsequent ticket / PR.

glennsarti:
:+1:  for merge from me.

Iristyle:
So there are still some issues here:

* Commit 172248bf80c23be135ada89b88095ee16fbd0a6f has the wrong file for `powershell_manager.rb`, so the diff is wigged out in fb40708
* We have that `> 3.8.5` exclusion everywhere, but no explanation in commit msg or in the upgrade message (which states `3.x` is fine)
* Iterations was added to `drain_pipe`, but the `upto` didn't get updated to use the variable.

Digging through to see if there are more.

lib/puppet/provider/exec/powershell.rb
```diff
@@ -25,17 +28,69 @@
...
+    Puppet.warning <<-UPGRADE
+The current Puppet version is outdated and uses a library that was
+previously necessary on the current Ruby verison to support a colored console.
+
+Unfortunately this library prevents the PowerShell module from using a shared
+PowerShell process to dramatically improve the performance of resource
+application.
+
+To enable these improvements, it is suggested to upgrade to any x64 version of
+Puppet (including 3.x), or to a Puppet version newer than 3.x.
```
Iristyle: This states `3.x`, but `powershell_manager` is only included on `> 3.8.5` above.

spec/integration/puppet_x/puppetlabs/powershell_manager_spec.rb
```diff
@@ -0,0 +1,144 @@
...
+
+      expect(result[:errormessage]).to eq(nil)
+      expect(result[:exitcode]).to eq(0)
+      expect(result[:stdout]).not_to eq(nil)
+    end
+
+    it "should return a JSON response with a timeout error if the execution timeout is exceeded" do
+      timeout_ms = 100
+      result = manager.execute('sleep 1', timeout_ms)
+      # TODO What is the real message now?
```
Iristyle: accidental TODO?

lib/puppet_x/puppetlabs/powershell_manager.rb
```diff
@@ -0,0 +1,208 @@
...
+      def write_stdin(input)
+        @stdin.puts(input)
+      rescue => e
+        msg = "Error writing STDIN / reading STDOUT: #{e}"
+        raise Puppet::Util::Windows::Error.new(msg)
+      end
+
+      def drain_pipe(pipe, iterations = 10)
+        output = []
+        0.upto(10) do
```
Iristyle: Whoops - should be `0.upto(iterations) do`

lib/puppet_x/puppetlabs/powershell_manager.rb
```diff
@@ -0,0 +1,208 @@
...
+        out = REXML::Document.new(out.gsub(/\n/,""))
+
+        hash = {}
+        REXML::XPath.each(out, "//Property").map { |x| [x.attributes["Name"], x.text] }.each do | prop |
+          hash[prop[0]] = prop[1]
+        end
+
+        {
+          :exitcode     => hash['exitcode'].to_i,
+          :errormessage => hash['errormessage'].nil? ? nil : Base64.decode64(hash['errormessage']),
```
Iristyle: `errormessage` and `stdout` are no longer ever nil ... but they can be empty strings.

lib/puppet/provider/exec/powershell.rb
```diff
@@ -1,4 +1,7 @@
 require 'puppet/provider/exec'
+if Puppet::Util::Platform.windows? &&  Puppet.version > '3.8.5'
```
Iristyle: I think this was a carryover from before `win32console_enabled?` came back... and I'm not sure it's correct.  Since we're using `win32console_enabled?` to guard against using this class, isn't the `3.8.5` part unnecessarily restrictive (and isn't `Puppet::Util::Platform.windows?` enough)?

spec/integration/puppet_x/puppetlabs/powershell_manager_spec.rb
```diff
@@ -0,0 +1,144 @@
+require 'spec_helper'
+require 'puppet/type'
+require 'puppet_x/puppetlabs/powershell_manager' if Puppet::Util::Platform.windows? &&  Puppet.version > '3.8.5'
```
Iristyle: As above, I think the `3.8.5` require is not necessary anymore.

spec/integration/puppet_x/puppetlabs/powershell_manager_spec.rb
```diff
@@ -0,0 +1,144 @@
...
+
+module PuppetX
+  module PowerShell
+    class PowerShellManager; end
+  end
+end
+
+describe PuppetX::PowerShell::PowerShellManager,
+  :if => Puppet::Util::Platform.windows? &&  Puppet.version > '3.8.5' do
```
Iristyle: I think the `3.8.5` check becomes a `win32console_enabled?` check

spec/unit/provider/exec/powershell_spec.rb
```diff
@@ -28,7 +28,7 @@
   }
 
   describe "#run" do
-    context "stubbed calls" do
+    context "stubbed calls", :if => Puppet.features.microsoft_windows? && Puppet.version <= '3.8.5' do
```
Iristyle: I think 3.8.5 check becomes `win32console_enabled?`

spec/unit/provider/exec/powershell_spec.rb
```diff
@@ -46,7 +46,7 @@
         provider.run_spec_override(command)
       end
 
-      it "should quote powershell.exe path", :if => Puppet.features.microsoft_windows? do
+      it "should quote powershell.exe path", :if => Puppet.features.microsoft_windows? && Puppet.version <= '3.8.5' do
```
Iristyle: I think 3.8.5 check becomes `win32console_enabled?`

Iristyle:
FYI - I'm working on fixing these bits up locally @jpogran - just ping me if you want to work on it with me.

PR #78, opened 2016-03-09
(MODULES-2962) Reuse PowerShell Session

Iristyle:
Supercedes #74 

Channeling @jpogran from previous PR:

> This PR is a port of the PowerShell session code from the DSC module. This code includes some significant speed and system utilization improvements.
> 
> In the prior version every call to exec started a brand new PowerShell process for each of the exec modes: command, onlyif and unless. This was costly in both time and system resources as PowerShell takes significant time to intialize and frequent process creations spike cpu on the target system.
> 
> In this version, a single PowerShell process is created that lives not only for all the modes of exec, but also for the entirety of the Puppet run. We still have a one time cost of time and system utilization for
> process creation for the entire run, but gain the ability to run consecutive commands.
> 
> A second improvement to this version is that we directly pass commands to PowerShell through shell redirection, and not through reading a file in like the last version. This avoids any of the parsing errors
> (MODULES-2634) we have seen. This also improves syntax checking, as we now validate the passed in commands using PowerShell itself before the command is run, allowing us to catch errors sooner and report them.
> 
> Backwards compatibility was an important requirement for this PR as well. The same behavior the old way of executing PowerShell must be maintained with this new way. This presented a problem as the long-lived session would keep all variables from Puppet run to Puppet run. Care was taken to reset the environment between runs by resetting both the PowerShell variables and Environment variables to original state each run. This also refreshed the PowerShell and Environment variables, allowing one to set global variables and have them take effect in the next run, just like the old behavior could.
> 
> While care was taken to run on all environments the old version did, we cannot use the new method with ruby 1.9.3. Ruby in this version used win32console, which breaks when stdout is redirected. This breaks the Puppet logging stream and halts the Puppet run. The code will detect if running on ruby 1.9.3 and use the old PowerShell invocation code, which while slower still allows the user to use this module.
> 
> This PR also removes support for Windows 2003, as Puppet itself has ceased support for Windows 2003.

### DIFF
--- a/.sync.yml
+++ b/.sync.yml
@@ -15,4 +15,4 @@ appveyor.yml:
     - PUPPET_GEM_VERSION: 3.0.0
       RUBY_VER: 193
   test_script:
-    - 'bundle exec rspec spec/unit -fd -b'
+    - 'bundle exec rspec spec/unit spec/integration -fd -b'
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -30,7 +30,7 @@ build: off
 test_script:
 - bundle exec puppet -V
 - ruby -v
-- bundle exec rspec spec/unit -fd -b
+- bundle exec rspec spec/unit spec/integration -fd -b
 notifications:
 - provider: Email
   to:
--- a/lib/puppet/provider/exec/powershell.rb
+++ b/lib/puppet/provider/exec/powershell.rb
@@ -1,4 +1,5 @@
 require 'puppet/provider/exec'
+require_relative '../../../puppet_x/puppetlabs/powershell_manager'
 
 Puppet::Type.type(:exec).provide :powershell, :parent => Puppet::Provider::Exec do
   confine :operatingsystem => :windows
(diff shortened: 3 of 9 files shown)


Iristyle:
@jpogran I think you meant `ruby 2.0.0x64 and puppet 3.8.5`? We don't support Ruby 2.0.0 x86 on Windows at all...

jpogran:
@Iristyle this line evaluates to 3.8.6 and 2.0.0x64 when run https://github.com/puppetlabs/puppetlabs-powershell/blob/master/appveyor.yml#L15-L16

Iristyle:
@jpogran I was referring to the second line above in your list referring to `3.8.5`

Iristyle:
Note that #82 is a follow-up PR for a couple of minor issues in the FFI rejiggering that was done as part of a4cd1f5.  Because AppVeyor wasn't running, and because I failed to do so as part of preparing those changes, we didn't get validation against Puppet 3.0 / Ruby 1.9.3 / x86.  The irony of course is that the code changes were part of a class declaration, for code that doesn't actually run on that platform...

PR #84, opened 2016-04-12
(MODULES-2634) Test try/catch in PowerShell provider

jpogran:
Due to the changes from MODULES-2962, ticket MODULES-2634 is fixed
as a side effect. Besides the module itself using try/catch semantics
in a templated PS1 file, it supports any kind of syntax or formatting..

This commit adds two tests to ensure that this is the case. One for
integration testing, the other for acceptance.


ferventcoder:
:+1:

PR #86, opened 2016-04-20
(MODULES-3137) Log PowerShell Streams

jpogran:
When we switched to a single PowerShell sessions using PowerShell
Runspaces, we lost the information from the Error, Debug, and Verbose
streams. By implementing a custom PowerShell Host, and the
PSHostUserInterface and PSHostRawUserInterface interfaces we capture
any information written to those streams

- [x] Check long running operation/streaming still work with all streams merged
- [x] Set PSHost version to 1.1
- [x] Test merged output '3>&2'
- [x] Fix wrapping behavior in text
- [x] Internationalization question on streams
- [x] Remove $output
- [x] C# Namespace - is it needed?
- [x] Check that write-progress is wanted
- [x] Check that user can use out-default

### DIFF
--- a/lib/puppet_x/templates/invoke_ps_command.erb
+++ b/lib/puppet_x/templates/invoke_ps_command.erb
@@ -1,14 +1,237 @@
 $hostSource = @"
 using System;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
 using System.Globalization;
+using System.Management.Automation;
 using System.Management.Automation.Host;
+using System.Security;
+using System.Text;
 using System.Threading;
 
-public class PuppetPSHost : PSHost
+namespace Puppet
 {
+  public class PuppetPSHostRawUserInterface : PSHostRawUserInterface
+  {
+    public PuppetPSHostRawUserInterface()
+    {
+      buffersize      = new Size(120, 120);
+      backgroundcolor = ConsoleColor.Black;
+      foregroundcolor = ConsoleColor.White;
+      cursorposition  = new Coordinates(0, 0);
+      cursorsize      = 1;
+    }
+
+    private ConsoleColor backgroundcolor;
+    public override ConsoleColor BackgroundColor
+    {
+      get { return backgroundcolor; }
+      set { backgroundcolor = value; }
+    }
+
+    private Size buffersize;
+    public override Size BufferSize
+    {
+      get { return buffersize; }
+      set { buffersize = value; }
+    }
+
+    private Coordinates cursorposition;
+    public override Coordinates CursorPosition
+    {
+      get { return cursorposition; }
+      set { cursorposition = value; }
+    }
+
+    private int cursorsize;
+    public override int CursorSize
+    {
+      get { return cursorsize; }
+      set { cursorsize = value; }
+    }
+
+    private ConsoleColor foregroundcolor;
+    public override ConsoleColor ForegroundColor
+    {
+      get { return foregroundcolor; }
+      set { foregroundcolor = value; }
+    }
+
+    private Coordinates windowposition;
+    public override Coordinates WindowPosition
+    {
+      get { return windowposition; }
+      set { windowposition = value; }
+    }
+
+    private Size windowsize;
+    public override Size WindowSize
+    {
+      get { return windowsize; }
+      set { windowsize = value; }
+    }
+
+    private string windowtitle;
+    public override string WindowTitle
+    {
+      get { return windowtitle; }
+      set { windowtitle = value; }
+    }
+
+    public override bool KeyAvailable
+    {
+        get { return false; }
+    }
+
+    public override Size MaxPhysicalWindowSize
+    {
+        get { return new Size(165, 66); }
+    }
+
+    public override Size MaxWindowSize
+    {
+        get { return new Size(165, 66); }
+    }
+
+    public override void FlushInputBuffer()
+    {
+      throw new NotImplementedException();
+    }
+
+    public override BufferCell[,] GetBufferContents(Rectangle rectangle)
+    {
+      throw new NotImplementedException();
+    }
+
+    public override KeyInfo ReadKey(ReadKeyOptions options)
+    {
+      throw new NotImplementedException();
+    }
+
+    public override void ScrollBufferContents(Rectangle source, Coordinates destination, Rectangle clip, BufferCell fill)
+    {
+      throw new NotImplementedException();
+    }
+
+    public override void SetBufferContents(Rectangle rectangle, BufferCell fill)
+    {
+      throw new NotImplementedException();
+    }
+
+    public override void SetBufferContents(Coordinates origin, BufferCell[,] contents)
+    {
+      throw new NotImplementedException();
+    }
+  }
+
+  public class PuppetPSHostUserInterface : PSHostUserInterface
+  {
+    private PuppetPSHostRawUserInterface _rawui;
+    private StringBuilder _sb;
+
+    public PuppetPSHostUserInterface()
+    {
+      _sb = new StringBuilder();
+    }
+
+    public override PSHostRawUserInterface RawUI
+    {
+      get
+      {
+        if ( _rawui == null){
+          _rawui = new PuppetPSHostRawUserInterface();
+        }
+        return _rawui;
+      }
+    }
+
+    public override void Write(ConsoleColor foregroundColor, ConsoleColor backgroundColor, string value)
+    {
+      _sb.Append(value);
+    }
+
+    public override void Write(string value)
+    {
+      _sb.Append(value);
+    }
+
+    public override void WriteDebugLine(string message)
+    {
+      _sb.AppendLine("DEBUG: " + message);
+    }
+
+    public override void WriteErrorLine(string value)
+    {
+      _sb.AppendLine(value);
+    }
+
+    public override void WriteLine(string value)
+    {
+      _sb.AppendLine(value);
+    }
+
+    public override void WriteVerboseLine(string message)
+    {
+      _sb.AppendLine("VERBOSE: " + message);
+    }
+
+    public override void WriteWarningLine(string message)
+    {
+      _sb.AppendLine("WARNING: " + message);
+    }
+
+    public override void WriteProgress(long sourceId, ProgressRecord record)
+    {
+    }
+
+    public string Output
+    {
+      get
+      {
+        string text = _sb.ToString();
+        _sb = new StringBuilder();
+        return text;
+      }
+    }
+
+    public override Dictionary<string, PSObject> Prompt(string caption, string message, Collection<FieldDescription> descriptions)
+    {
+      throw new NotImplementedException();
+    }
+
+    public override int PromptForChoice(string caption, string message, Collection<ChoiceDescription> choices, int defaultChoice)
+    {
+      throw new NotImplementedException();
+    }
+
+    public override PSCredential PromptForCredential(string caption, string message, string userName, string targetName)
+    {
+      throw new NotImplementedException();
+    }
+
+    public override PSCredential PromptForCredential(string caption, string message, string userName, string targetName, PSCredentialTypes allowedCredentialTypes, PSCredentialUIOptions options)
+    {
+      throw new NotImplementedException();
+    }
+
+    public override string ReadLine()
+    {
+      throw new NotImplementedException();
+    }
+
+    public override SecureString ReadLineAsSecureString()
+    {
+      throw new NotImplementedException();
+    }
+  }
+
+  public class PuppetPSHost : PSHost
+  {
     private Guid _hostId = Guid.NewGuid();
     private bool shouldExit;
     private int exitCode;
+
+    private readonly PuppetPSHostUserInterface _ui = new PuppetPSHostUserInterface();
 
     public PuppetPSHost () {}
 
@@ -22,8 +245,11 @@ public class PuppetPSHost : PSHost
 
     public override Guid InstanceId { get { return _hostId; } }
     public override string Name { get { return "PuppetPSHost"; } }
-    public override Version Version { get { return new Version(1, 0); } }
-    public override PSHostUserInterface UI { get { return null; } }
+    public override Version Version { get { return new Version(1, 1); } }
+    public override PSHostUserInterface UI
+    {
+      get { return _ui; }
+    }
     public override CultureInfo CurrentCulture
     {
         get { return Thread.CurrentThread.CurrentCulture; }
@@ -43,6 +269,7 @@ public class PuppetPSHost : PSHost
       this.shouldExit = true;
       this.exitCode = exitCode;
     }
+  }
 }
 "@
 
@@ -131,7 +358,7 @@ if ($runspace -eq $null){
     $sessionState = [System.Management.Automation.Runspaces.InitialSessionState]::CreateDefault()
   }
 
-  $puppetPSHost = New-Object PuppetPSHost
+  $puppetPSHost = New-Object Puppet.PuppetPSHost
   $runspace = [System.Management.Automation.Runspaces.RunspaceFactory]::CreateRunspace($puppetPSHost, $sessionState)
   $runspace.Open()
 }
@@ -151,11 +378,12 @@ try
 
   $ps = [System.Management.Automation.PowerShell]::Create()
   $ps.Runspace = $runspace
-
   [Void]$ps.AddScript($ourFunctions)
   $ps.Invoke()
 
+
   if(!$environmentVariables){
+    $ps.Commands.Clear()
     $environmentVariables = $ps.AddCommand('Get-ProcessEnvironmentVariables').Invoke()
   }
 
@@ -163,26 +391,46 @@ try
     if(!$psVariables){
       $psVariables = $ps.AddScript('Get-Variable').Invoke()
     }
+
+    $ps.Commands.Clear()
     [void]$ps.AddScript('Get-Variable -Scope Global | Remove-Variable -Force -ErrorAction SilentlyContinue -WarningAction SilentlyContinue')
     $ps.Invoke()
+
+    $ps.Commands.Clear()
     [void]$ps.AddCommand('Reset-ProcessPowerShellVariables').AddParameter('psVariables', $psVariables)
     $ps.Invoke()
   }
 
+  $ps.Commands.Clear()
   [Void]$ps.AddCommand('Reset-ProcessEnvironmentVariables').AddParameter('processVars', $environmentVariables)
   $ps.Invoke()
 
+  # we clear the commands before each new command
+  # to avoid command pollution
+  $ps.Commands.Clear()
   [Void]$ps.AddScript($powershell_code)
+
+  # out-default and MergeMyResults takes all output streams
+  # and writes it to the PSHost we create
+  # this needs to be the last thing executed
+  [void]$ps.AddCommand("out-default");
+  if($PSVersionTable.PSVersion -le [Version]'2.0'){
+    $ps.Commands.Commands[0].MergeMyResults([System.Management.Automation.Runspaces.PipelineResultTypes]::Error, [System.Management.Automation.Runspaces.PipelineResultTypes]::Output);
+  }else{
+    $ps.Commands.Commands[0].MergeMyResults([System.Management.Automation.Runspaces.PipelineResultTypes]::All, [System.Management.Automation.Runspaces.PipelineResultTypes]::Output);
+  }
   $asyncResult = $ps.BeginInvoke()
 
   if (!$asyncResult.AsyncWaitHandle.WaitOne(<%= timeout_ms %>, $false)){
     throw "Catastrophic failure: PowerShell module timeout (<%= timeout_ms %> ms) exceeded while executing"
   }
 
-  $output = $ps.EndInvoke($asyncResult)
-  $output = $output | Out-String
+  $ps.EndInvoke($asyncResult)
 
-  New-XmlResult -exitcode $puppetPSHost.Exitcode -output $output -errormessage $null
+  [Puppet.PuppetPSHostUserInterface]$ui = $puppetPSHost.UI
+  [string]$text = $ui.Output
+
+  New-XmlResult -exitcode $puppetPSHost.Exitcode -output $text -errormessage $null
 }
 catch
 {
--- a/spec/integration/puppet_x/puppetlabs/powershell_manager_spec.rb
+++ b/spec/integration/puppet_x/puppetlabs/powershell_manager_spec.rb
@@ -141,4 +141,71 @@ $bytes_in_k = (1024 * 64) + 1
     end
   end
 
+  describe "when output is written to a PowerShell Stream" do
+    it "should collect anything written to verbose stream" do
+      result = manager.execute('$VerbosePreference = "Continue";Write-Verbose "Hello"')
+
+      expect(result[:stdout]).to eq("VERBOSE: Hello\r\n")
+      expect(result[:exitcode]).to eq(0)
+    end
+
+    it "should collect anything written to debug stream" do
+      result = manager.execute('$debugPreference = "Continue";Write-debug "Hello"')
+
+      expect(result[:stdout]).to eq("DEBUG: Hello\r\n")
+      expect(result[:exitcode]).to eq(0)
+    end
+
+    it "should collect anything written to Warning stream" do
+      result = manager.execute('Write-Warning "Hello"')
+
+      expect(result[:stdout]).to eq("WARNING: Hello\r\n")
+      expect(result[:exitcode]).to eq(0)
+    end
+
+    it "should collect anything written to Error stream" do
+      result = manager.execute('Write-Error "Hello"')
+
+      expect(result[:stdout]).to eq("Write-Error \"Hello\" : Hello\r\n    + CategoryInfo          : NotSpecified: (:) [Write-Error], WriteErrorException\r\n    + FullyQualifiedErrorId : Microsoft.PowerShell.Commands.WriteErrorException\r\n \r\n")
+      expect(result[:exitcode]).to eq(0)
+    end
+
+    it "should handle a Write-Error in the middle of code" do
+      result = manager.execute('ls;Write-Error "Hello";ps')
+
+      expect(result[:stdout]).not_to eq(nil)
+      expect(result[:exitcode]).to eq(0)
+    end
+
+    it "should handle a Out-Default in the user code" do
+      result = manager.execute('\'foo\' | Out-Default')
+
+      expect(result[:stdout]).to eq("foo\r\n")
+      expect(result[:exitcode]).to eq(0)
+    end
+
+    it "should handle lots of output from user code" do
+      result = manager.execute('1..1000 | %{ (65..90) + (97..122) | Get-Random -Count 5 | % {[char]$_} }')
+
+      expect(result[:stdout]).not_to eq(nil)
+      expect(result[:exitcode]).to eq(0)
+    end
+
+    it "should handle a larger return of output from user code" do
+      result = manager.execute('1..1000 | %{ (65..90) + (97..122) | Get-Random -Count 5 | % {[char]$_} } | %{ $f="" } { $f+=$_ } {$f }')
+
+      expect(result[:stdout]).not_to eq(nil)
+      expect(result[:exitcode]).to eq(0)
+    end
+
+    it "should handle shell redirection" do
+      # the test here is to ensure that this doesn't break. because we merge the streams regardless
+      # the opposite of this test shows the same thing
+      result = manager.execute('function test-warning{ ps;write-warning \'foo\' }; test-warning 3>&1')
+
+      expect(result[:stdout]).not_to eq(nil)
+      expect(result[:exitcode]).to eq(0)
+    end
+  end
+
 end


jpogran:
Appveyor will fail until #87 is merged

lib/puppet_x/templates/invoke_ps_command.erb
```diff
@@ -173,14 +369,25 @@ try
   $ps.Invoke()
 
   [Void]$ps.AddScript($powershell_code)
+  $ps.Commands.Commands[0].MergeMyResults(
```
jpogran: This, combined with the PSHostUserInterface and PSHostRawUserInterface, will output Info, Debug, Verbose, Warning to 'stdout'. The risk here is that we don't get PowerShell objects back, we get a string with output from all Streams merged together. Since we are looking for this, this is likely not a problem.

lib/puppet_x/templates/invoke_ps_command.erb
```diff
@@ -1,15 +1,207 @@
...
+      }
+    }
+
+    public string Output
+    {
+      get
+      {
+        var text = _sb.ToString();
+        _sb.Clear();
+        return text;
```
jpogran: So, we are running inside a PowerShell runspace, with PSHostRawUserInterface and PowerShellHostUserInterface implemented. Most examples put a System.Console in here and call it a day. However, System.Console doesn't work because we don't have a true console at this point to write to and anything we send to doesn't get captured by the PowerShell runspace..

In order to get the output, I have instead all the Write* methods write to a StringBuilder which then returns the string after execution is done. Stringbuilder can be a memory hog and over allocate, but this is destroyed when the entire Puppet run is done so I'm not sure if I'm overthinking being worried about it here.

lib/puppet_x/templates/invoke_ps_command.erb
```diff
@@ -173,14 +369,25 @@ try
...
 
   $output = $ps.EndInvoke($asyncResult)
-  $output = $output | Out-String
+  #$output = $output | Out-String
+  #if($ps.HasErrors){
+  #  $ps.Streams.Error.ReadAll() | %{
+  #    $puppetPSHost.UI.WriteLine($_.ToString())
+  #  }
+  #}
+  $output = [string]($puppetPSHost.UI.Output)
```
jpogran: Since we are getting out output from the UI we don't actually need the output returned from the Invoke() call, as that will always be empty because of the host we implemented

lib/puppet_x/templates/invoke_ps_command.erb
ferventcoder: System.Console is actually harmful when using redirection - it was a bug I ran into with choco recently. keeping it in a string builder for return works. 👍

lib/puppet_x/templates/invoke_ps_command.erb
```diff
@@ -151,38 +347,58 @@ try
...
     $ps.Invoke()
   }
 
+  $ps = [System.Management.Automation.PowerShell]::Create()
+  $ps.Runspace = $runspace
   [Void]$ps.AddCommand('Reset-ProcessEnvironmentVariables').AddParameter('processVars', $environmentVariables)
   $ps.Invoke()
 
+  $ps = [System.Management.Automation.PowerShell]::Create()
+  $ps.Runspace = $runspace
```
ferventcoder: The the multiple creates?

lib/puppet_x/templates/invoke_ps_command.erb
ferventcoder: Have you thought about keeping a string builder per log stream? That way you don't always show the debug messages?

lib/puppet_x/templates/invoke_ps_command.erb
```diff
@@ -1,15 +1,207 @@
...
+        get { return Console.KeyAvailable; }
+    }
+    public override Size MaxPhysicalWindowSize
+    {
+        get { return new Size(Console.LargestWindowWidth, Console.LargestWindowHeight); }
+    }
+
+    public override Size MaxWindowSize
+    {
+        get { return new Size(Console.LargestWindowWidth, Console.LargestWindowHeight); }
```
ferventcoder: I'd be careful on these. They can cause issues.

lib/puppet_x/templates/invoke_ps_command.erb
```diff
@@ -1,15 +1,207 @@
...
+    public override Coordinates CursorPosition { get; set; }
+    public override int CursorSize { get; set; }
+    public override ConsoleColor ForegroundColor { get; set; }
+    public override Coordinates WindowPosition { get; set; }
+    public override Size WindowSize { get; set; }
+    public override string WindowTitle { get; set; }
+
+    public override bool KeyAvailable
+    {
+        get { return Console.KeyAvailable; }
```
ferventcoder: I think you are going to want https://github.com/chocolatey/choco/blob/439378526bd0c835723c34312c9fad6e211180c7/src/chocolatey/infrastructure/adapters/Console.cs#L142-L150

lib/puppet_x/templates/invoke_ps_command.erb
ferventcoder: https://github.com/chocolatey/choco/blob/439378526bd0c835723c34312c9fad6e211180c7/src/chocolatey/infrastructure/adapters/Console.cs#L166-L174

lib/puppet_x/templates/invoke_ps_command.erb
```diff
@@ -1,15 +1,207 @@
...
-public class PuppetPSHost : PSHost
+namespace Puppet
 {
+  public class PowerShellRawUserInterface : PSHostRawUserInterface
+  {
+    public PowerShellRawUserInterface()
+    {
+      BufferSize = new Size(80, 25);
+      BackgroundColor = ConsoleColor.Black;
+      ForegroundColor = ConsoleColor.White;
```
ferventcoder: Not sure you need to fully qualify this, but isn't it `[System.ConsoleColor]::White`?

ferventcoder: Wait, c# - nevermind!!!

lib/puppet_x/templates/invoke_ps_command.erb
jpogran: I could get Verbose, Debug and Progress streams with our current code, but not the Error stream (see our ticket for explanation). After a long advanced debugging session (read: comment part of code, run, then uncomment, repeat) I found that if we didn't create a new PowerShell object after the call to Reset-ProcessEnvironmentVariables, I would loose the formatting for any Error object and instead just get the message, not the fully formatted string.

Last time i used C# and a PowerShell host, I used a new PowerShell.Create object for each command, but can't point you to a specific documentation site stating the need. The Runspace object is where the state is, the PowerShell object represents the pipeline part, so this makes sense to me. I wanted to make a helper function to reduce the double lines everywhere, but saw the extra function as potentially more confusing down the line.

lib/puppet_x/templates/invoke_ps_command.erb
jpogran: This entire class was added only to get the default PowerShell formatting to work, its not usually required by default. We don't actually use the console, just attempt to access the constants.

lib/puppet_x/templates/invoke_ps_command.erb
jpogran: We don't always show debug messages

lib/puppet_x/templates/invoke_ps_command.erb
ferventcoder: Would you mind adding some comments as to that in the code to reduce the confusion for when we've forgotten this conversation later?

lib/puppet_x/templates/invoke_ps_command.erb
```diff
@@ -1,15 +1,207 @@
...
+      CursorPosition = new Coordinates(0, 0);
+      CursorSize = 1;
+    }
+
+    public override ConsoleColor BackgroundColor { get; set; }
```
ferventcoder: Puppet.PowerShellRawUserInterface.BackgroundColor.get must declare a body because it is not marked as asbtract or extern

ferventcoder: I forget, compatible with .NET 2.x? You can't do auto properties.

lib/puppet_x/templates/invoke_ps_command.erb
```diff
@@ -23,7 +215,10 @@ public class PuppetPSHost : PSHost
     public override Guid InstanceId { get { return _hostId; } }
     public override string Name { get { return "PuppetPSHost"; } }
     public override Version Version { get { return new Version(1, 0); } }
```
ferventcoder: There may be some things that actually check `$host.Version` - not saying they should, I'm just saying they are likely to. I wonder how they would react to 1.0?

lib/puppet_x/templates/invoke_ps_command.erb
ferventcoder: @jpogran ping when you get back to this.

lib/puppet_x/templates/invoke_ps_command.erb
glennsarti: What version would we use? Isn't this the same issue as getting the Host version when over WinRM? or Chocolatey?

jpogran: The `PSHost` version indicates the version of the hosting application, not PowerShell. The `ConsoleHost` host and the `ISE` host share the same number as the version of PowerShell installed, but as @glennsarti points out `WinRM` and other hosts like `Chocolately` don't. The Host name and version are supposed to allow you to make the decision as a module author whether or not you support running on that host. I'd argue anyone specifying a specific host `name` and `version` know what they are doing enough to make the call.

The alternative is to mimic the ConsoleHost host name and version, which is misleading and may lead to unintended consequences.

lib/puppet_x/templates/invoke_ps_command.erb
jpogran: Last note, if we did separate StringBuilders for each stream, it would be out of order when we finally output the result

lib/puppet_x/templates/invoke_ps_command.erb
Iristyle: I agree with @jpogran here, and I think that we should not be matching version to PS version, etc.  I've had to introspect hosts in utility code to detect features (i.e. WinRM doesn't do transcription) - so not mirroring the name / version of other hosts is the way to go IMHO.  I would say we should just use `base.Version`, except the field is `abstract` in [PSHost](https://msdn.microsoft.com/en-us/library/system.management.automation.host.pshost.version\(v=vs.85\).aspx).

I will say though that we should probably be bumping the version from the last `1.0` version in the module, to something newer like `1.1`, since we've added new features to the class (even though they're not shipped yet).

spec/integration/puppet_x/puppetlabs/powershell_manager_spec.rb
```diff
@@ -141,4 +141,48 @@ class PowerShellManager; end
     end
   end
 
+  describe "when output is written to a PowerShell Stream" do
```
Iristyle: Do we want to test a script that uses merged output like `3>&2`?  (or anything that doesn't merge it to `1`)

lib/puppet_x/templates/invoke_ps_command.erb
```diff
@@ -1,15 +1,255 @@
...
+    {
+      throw new NotImplementedException();
+    }
+
+    public override string ReadLine()
+    {
+      throw new NotImplementedException();
+    }
+
+    public override System.Security.SecureString ReadLineAsSecureString()
```
Iristyle: Nit - we're already `using System.Security`, so no need to fully qualify `SecureString`

lib/puppet_x/templates/invoke_ps_command.erb
Iristyle: Using public properties to set private variables in a C# class constructor is not idiomatic.  Prefer to set the backing fields directly - i.e.

```csharp
buffersize = new Size(80, 25);
```

lib/puppet_x/templates/invoke_ps_command.erb
```diff
@@ -1,15 +1,255 @@
...
+    }
+
+    public override bool KeyAvailable
+    {
+        get { return false; }
+    }
+
+    public override Size MaxPhysicalWindowSize
+    {
+        get { return new Size(165, 66); }
```
Iristyle: How was this value determined?  Does this change how strings are wrapped in any way?

lib/puppet_x/templates/invoke_ps_command.erb
```diff
@@ -1,15 +1,255 @@
...
+    }
+
+    public override void WriteVerboseLine(string message)
+    {
+      _sb.AppendLine("VERBOSE: " + message);
+    }
+
+    public override void WriteWarningLine(string message)
+    {
+      _sb.AppendLine("WARNING: " + message);
```
Iristyle: These prefixes are probably fine, but I wonder if there's a way that we can retrieve these values from constants somewhere?  I also wonder if these values change based on your current locale or not / should they?

Iristyle: Also - does this mean that we lose colored output?

lib/puppet_x/templates/invoke_ps_command.erb
Iristyle: Yeah, multiple StringBuilders would be a real pain, since you would need to replay them in order - so I'm not in favor of that.

Do we have any behavioral changes with respect to output streaming / colorization?  Before, we weren't collecting any of the stream output up... but now we are. Does this mean that the output behavior for long-running PS operations has changed?

i.e. what happens if you have a PS script that runs in a one second incremented loop over 20 seconds, and writes a message during each loop... what does that look like executing the Puppet run at the local command prompt in the v1 module vs the current code vs this new code?

lib/puppet_x/templates/invoke_ps_command.erb
```diff
@@ -151,38 +395,53 @@ try
...
+  # this needs to be the last thing executed
+  [void]$ps.AddCommand("out-default");
+  $ps.Commands.Commands[0].MergeMyResults([System.Management.Automation.Runspaces.PipelineResultTypes]::All, [System.Management.Automation.Runspaces.PipelineResultTypes]::Output);
   $asyncResult = $ps.BeginInvoke()
 
   if (!$asyncResult.AsyncWaitHandle.WaitOne(<%= timeout_ms %>, $false)){
     throw "Catastrophic failure: PowerShell module timeout (<%= timeout_ms %> ms) exceeded while executing"
   }
 
   $output = $ps.EndInvoke($asyncResult)
```
Iristyle: What's in `$output` now?  The variable is no longer used after calling `EndInvoke` now...

lib/puppet_x/templates/invoke_ps_command.erb
```diff
@@ -1,15 +1,255 @@
...
+using System.Management.Automation;
 using System.Management.Automation.Host;
+using System.Security;
+using System.Text;
 using System.Threading;
 
-public class PuppetPSHost : PSHost
+namespace Puppet
 {
+  public class PowerShellRawUserInterface : PSHostRawUserInterface
```
Iristyle: Minor - why add the namespace?  Why not just keep the class name consistent to avoid collisions (or is there some concern over polluting namespace)?

* `PuppetPSHost` exists derived from `PSHost` ... would think then that we use:
* `PuppetPSHostRawUserInterface` derived from `PSHostRawUserInterface`

lib/puppet_x/templates/invoke_ps_command.erb
```diff
@@ -1,15 +1,255 @@
...
+    public override void WriteWarningLine(string message)
+    {
+      _sb.AppendLine("WARNING: " + message);
+    }
+
+    private bool hasLoggedStartProgress = false;
+    private bool hasLoggedFinalProgress = false;
+    public override void WriteProgress(long sourceId, ProgressRecord record)
+    {
+      if(record.PercentComplete == -1) return;
```
Iristyle: Style nit - spaces after `if`

lib/puppet_x/templates/invoke_ps_command.erb
```diff
@@ -1,15 +1,255 @@
...
+      if(hasLoggedFinalProgress) return;
+      if(!hasLoggedStartProgress)
+      {
+        hasLoggedStartProgress = true;
+        _sb.AppendLine(record.Activity.ToString());
+      }
+
+      // http://stackoverflow.com/a/888569/18475
+      var msg = string.Format("\rProgress: {0}% - {1}", record.PercentComplete.ToString(), record.StatusDescription.PadRight(50, ' ') );
+      _sb.Append(msg);
```
Iristyle: What does this actually look like in a log?  Seems like it could generate a ton of output when progress logs in `1%` increments (i.e. that would generate 100 lines of mostly useless text).  Do we care?

lib/puppet_x/templates/invoke_ps_command.erb
```diff
@@ -151,38 +395,53 @@ try
...
   [Void]$ps.AddCommand('Reset-ProcessEnvironmentVariables').AddParameter('processVars', $environmentVariables)
   $ps.Invoke()
 
+  $ps.Commands.Clear()
   [Void]$ps.AddScript($powershell_code)
+
+  # out-default and MergeMyResults takes all output streams
+  # and writes it to the PSHost we create
+  # this needs to be the last thing executed
+  [void]$ps.AddCommand("out-default");
```
Iristyle: What happens if a user calls this in their script?

lib/puppet_x/templates/invoke_ps_command.erb
jpogran: They were originally public auto properties that were converted to private backing variables to support .NET 2.0.

lib/puppet_x/templates/invoke_ps_command.erb
jpogran: Taking this even further, we will have PSHost version 1.1, in Puppet PowerShell module 2.0.0, with PowersShell versions 2.0, 3.0, 4.0, and 5.0. Should the PSHost version match the module version to reduce the amount of differences?

lib/puppet_x/templates/invoke_ps_command.erb
jpogran: A comment from @ferventcoder lead to hardcoding values here. Using constants from the Console class was advised against as dangerous by him, so the values from these constants were used.

This affects when and where the strings are wrapped. We are using the defaults so this mimics existing behavior.

lib/puppet_x/templates/invoke_ps_command.erb
jpogran: Nothing is returned on Invoke, as we told PS to merge all results with `MergeMyResults`

lib/puppet_x/templates/invoke_ps_command.erb
jpogran: Namespace was added from personal C# habits. Thought it would be better to make it namspaced under us to avoid possible (unlikely collisions)

lib/puppet_x/templates/invoke_ps_command.erb
jpogran: This was lifted from Choco. I believe the intent was to mimic the existing progress as much as possible. The ticket requested all output, so I implemented it. If we don't want it, thats fine.

lib/puppet_x/templates/invoke_ps_command.erb
jpogran: You can call out-default a million times and it won't change anything.

lib/puppet_x/templates/invoke_ps_command.erb
Iristyle: I'm seeing some weird wrapping behavior that wasn't present in the current v1 module.

Stable - i.e. v1

Current master - i.e. v2 + this PR

lib/puppet_x/templates/invoke_ps_command.erb
Iristyle: As a follow up, I've determined my concerns about colorization were not warranted.  I am still a bit curious about the streaming bits, which should be easy to rig up a quick test for.

Buffering lots of output in our own stringbuilders could be problematic under certain circumstances as you've noted @jpogran

lib/puppet_x/templates/invoke_ps_command.erb
Iristyle: @jpogran the only problem is then we have to bump the version each time the module releases, which I don't think is practical (especially if nothing changes in our PSHost - which should be the common case).

lib/puppet_x/templates/invoke_ps_command.erb
Iristyle: Ok, so that means we don't need to assign to `$output` anymore.

lib/puppet_x/templates/invoke_ps_command.erb
Iristyle: 👍  - just making sure that's the case

lib/puppet_x/templates/invoke_ps_command.erb
glennsarti: I like the idea of matching the PSHost Version to the module version but I can see @Iristyle point of view.  If it could be easily automated during the module build process then maybe...

lib/puppet_x/templates/invoke_ps_command.erb
jpogran: Test added to ensure truthiness

lib/puppet_x/templates/invoke_ps_command.erb
jpogran: Normal output:

```
PS > bundle exec puppet apply --modulepath C:\src\puppetlabs\modules erro
r.pp
DL is deprecated, please use Fiddle
Notice: Compiled catalog for win2012r2x64 in environment production in 0.24 seconds
Notice: /Stage[main]/Main/Exec[write-progress]/returns: Search in Progress
Progress: 100% - 100% Complete:
Notice: /Stage[main]/Main/Exec[write-progress]/returns: executed successfully
Notice: Applied catalog in 1.09 seconds
```

Debug output shows a very large XML output, but as the link in the code explains it resolves to one line.

@puppetlabs/windows do we want to ignore the progress stream in this PR?

lib/puppet_x/templates/invoke_ps_command.erb
jpogran: Renamed as suggested, but kept in namespace. Just feel odd having three classes not inside a namespace.

lib/puppet_x/templates/invoke_ps_command.erb
jpogran: @Iristyle Added two new tests. One generates long list of text, the other generates a large block of text. I fear doing anything like `Start-Sleep` will slow down unit tests too much. Thoughts?

spec/integration/puppet_x/puppetlabs/powershell_manager_spec.rb
jpogran: Could you suggest a different example? `3>&2` doesn't work

lib/puppet_x/templates/invoke_ps_command.erb
jpogran: Fixed by adjusting buffersize

lib/puppet_x/templates/invoke_ps_command.erb
jpogran: I think the concern about internationalization is valid, but can't be addressed in this ticket. While there are built-in ways to do i18n in powershell and C#, I don't see a way to use them here. The ways I know of require us to provide the translations in hashes or resx files. It's going to require some research to figure out our path forward.

spec/integration/puppet_x/puppetlabs/powershell_manager_spec.rb
Iristyle: It looks like PowerShell specially parses requests to merge streams (and must have special internal handling), and doesn't allow arbitrary stream / handle merging as I've suggested (based on what `cmd.exe` allows).  I think as long as something like `3>&1` works OK, then we're probably fine.

lib/puppet_x/templates/invoke_ps_command.erb
Iristyle: I did a little spelunking to dig the PS constants out.  They are currently hardcoded, and not internationalized within PowerShell itself, and unfortunately they're internal, so accessing them requires reflection - but it's not so bad.

```csharp
Type strings = Assembly.GetAssembly(typeof(System.Management.Automation.PowerShell)).GetType("InternalHostUserInterfaceStrings");
var infoFormat = strings.GetProperty("InformationFormatString", BindingFlags.Static | BindingFlags.NonPublic);
var warningFormat = strings.GetProperty("WarningFormatString", BindingFlags.Static | BindingFlags.NonPublic);
var verboseFormat = strings.GetProperty("VerboseFormatString", BindingFlags.Static | BindingFlags.NonPublic);
var debugFormat = strings.GetProperty("DebugFormatString", BindingFlags.Static | BindingFlags.NonPublic);
```

lib/puppet_x/templates/invoke_ps_command.erb
glennsarti: The v1 Powershell Module behaviour was to NOT write the progress to the stdout I'm happy with silently ignoring Write-Progress

lib/puppet_x/templates/invoke_ps_command.erb
Iristyle: FYI - `MshCommandRuntime` is the class where PS itself uses these string formats.

lib/puppet_x/templates/invoke_ps_command.erb
Iristyle: Yeah, if we didn't do it before.. and no one has asked about it, then perhaps we're OK without it.  

I'm fine either way to be honest - if anyone feels really strongly that we should / shouldn't do it and has compelling reasons, I'd be inclined to align with them.

@ferventcoder you want to add your 2c?

Iristyle:
Those AppVeyor failures are ... interesting.

jpogran:
The failures are interesting. Inspecting using pry shows that in ruby 200-x64 stderr shows the following errors:

```
=> {:exitcode=>1,
 :errormessage=>"You cannot call a method on a null-valued expression.\r\n",
 :stdout=>nil,
 :stderr=>
  [["Add-Type : c:\\Users\\vagrant\\AppData\\Local\\Temp\\tdwcpp0j.0.cs(14) : Method must have a return type\n",
    "c:\\Users\\vagrant\\AppData\\Local\\Temp\\tdwcpp0j.0.cs(13) :   {\n",
    "c:\\Users\\vagrant\\AppData\\Local\\Temp\\tdwcpp0j.0.cs(14) : >>>     public PowerShellRawUserInterface()\n",
    "c:\\Users\\vagrant\\AppData\\Local\\Temp\\tdwcpp0j.0.cs(15) :     {\n",
    "At line:1 char:1\n",
    "+ Add-Type -TypeDefinition $hostSource -Language CSharp\n",
    "+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~\n",
    "    + CategoryInfo          : InvalidData: (c:\\Users\\vagran...e a return type:CompilerError) [Add-Type], Exceptio
n\n",
    "    + FullyQualifiedErrorId : SOURCE_CODE_ERROR,Microsoft.PowerShell.Commands.AddTypeCommand\n",
    " \n",
    "Add-Type : c:\\Users\\vagrant\\AppData\\Local\\Temp\\tdwcpp0j.0.cs(111) : Method must have a return type\n"],
   ["c:\\Users\\vagrant\\AppData\\Local\\Temp\\tdwcpp0j.0.cs(110) :     private StringBuilder _sb;\n",
    "c:\\Users\\vagrant\\AppData\\Local\\Temp\\tdwcpp0j.0.cs(111) : >>>     public PowerShellHostUserInterface()\n",
    "c:\\Users\\vagrant\\AppData\\Local\\Temp\\tdwcpp0j.0.cs(112) :     {\n",
    "At line:1 char:1\n",
    "+ Add-Type -TypeDefinition $hostSource -Language CSharp\n",
    "+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~\n",
    "    + CategoryInfo          : InvalidData: (c:\\Users\\vagran...e a return type:CompilerError) [Add-Type], Exceptio
n\n",
    "    + FullyQualifiedErrorId : SOURCE_CODE_ERROR,Microsoft.PowerShell.Commands.AddTypeCommand\n",
    " \n",
    "Add-Type : Cannot add type. Compilation errors occurred.\n",
    "At line:1 char:1\n",
    "+ Add-Type -TypeDefinition $hostSource -Language CSharp\n",
    "+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~\n",
    "    + CategoryInfo          : InvalidData: (:) [Add-Type], InvalidOperationException\n",
    "    + FullyQualifiedErrorId : COMPILER_ERRORS,Microsoft.PowerShell.Commands.AddTypeCommand\n",
    " \n",
    "New-Object : Cannot find type [Puppet.PuppetPSHost]: verify that the assembly containing this type is loaded.\n",
    "At line:8 char:19\n",
    "+   $puppetPSHost = New-Object Puppet.PuppetPSHost\n",
    "+                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~\n",
    "    + CategoryInfo          : InvalidType: (:) [New-Object], PSArgumentException\n",
    "    + FullyQualifiedErrorId : TypeNotFound,Microsoft.PowerShell.Commands.NewObjectCommand\n",
    " \n",
    "Exception calling \"CreateRunspace\" with \"2\" argument(s): \"Cannot process argument because the value of argumen
t \"host\" \n",
    "is null. Change the value of argument \"host\" to a non-null value.\"\n",
    "At line:9 char:3\n",
    "+   $runspace = [System.Management.Automation.Runspaces.RunspaceFactory]::CreateRu ...\n",
    "+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~\n",
    "    + CategoryInfo          : NotSpecified: (:) [], MethodInvocationException\n",

 "    + FullyQualifiedErrorId : PSArgumentNullException\n",
 " \n",
 "You cannot call a method on a null-valued expression.\n",
 "At line:10 char:3\n",
 "+   $runspace.Open()\n",
 "+   ~~~~~~~~~~~~~~~~\n",
 "    + CategoryInfo          : InvalidOperation: (:) [], RuntimeException\n",
 "    + FullyQualifiedErrorId : InvokeMethodOnNull\n",
 " \n"]]}
```

Since this works in newer and older versions of ruby and puppet, this leads me to believe there is something different in this ruby version that is reading erb templates, causing the string we are sending to the C# compiler to not recognize a valid constructor call.

jpogran:
Nothing to see here, just a badly done class rename.

lib/puppet_x/templates/invoke_ps_command.erb
jpogran: This will be addressed in MODULES-3346

lib/puppet_x/templates/invoke_ps_command.erb
Iristyle: My comments about the streaming were about how the output is processed / buffered / sent through to Puppet - but as we discussed this AM, I don't think that the process has changed here.

glennsarti:
I'm happy for merge 👍

lib/puppet_x/templates/invoke_ps_command.erb
```diff
@@ -1,15 +1,238 @@
...
+    public override void WriteProgress(long sourceId, ProgressRecord record)
+    {
+    }
+
+    public string Output
+    {
+      get
+      {
+        var text = _sb.ToString();
+        _sb.Clear();
```
ferventcoder: We talked about these two - will need fixed prior to merge.

lib/puppet_x/templates/invoke_ps_command.erb
```diff
@@ -151,38 +378,55 @@ try
...
+  # we clear the commands before each new command
+  # to avoid command pollution
+  $ps.Commands.Clear()
   [Void]$ps.AddScript($powershell_code)
+
+  # out-default and MergeMyResults takes all output streams
+  # and writes it to the PSHost we create
+  # this needs to be the last thing executed
+  [void]$ps.AddCommand("out-default");
+  $ps.Commands.Commands[0].MergeMyResults([System.Management.Automation.Runspaces.PipelineResultTypes]::All, [System.Management.Automation.Runspaces.PipelineResultTypes]::Output);
```
ferventcoder: This also needs changed prior to merge (as we found).

ferventcoder:
With those two changes, I'm very happy with this.

ferventcoder:
I'm happy with this. If it hasn't had any other comments by @puppetlabs/windows by COB today, I will merge it.

PR #87, opened 2016-04-20
(MODULES-3280) Remove Verbose Environment Variable Handling

jpogran:
When the function to reset the Environment Variables for the single
session PowerShell process was added, the Verbose paramater was mistakenly
left on the Remove-Item and Set-Item cmdlets. This causes a Verbose
message to be written every time the provider is executed, polluting the
Puppet Debug stream. This is only an issue when MODULES-3137 is
implemented as it enables logging Verbose statements.

This removes the Verbose parameters.


jpogran:
#86 needs this to stop it's Appveyor tests from failing

PR #91, opened 2016-04-26
(maint) Update module for new Puppet branding

glennsarti:
Updated the module with the new branding and updated the license date
Updated the CONTRIBUTING file and removed dead links

[ci skip]

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -186,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright 2013 Puppet Labs
+   Copyright 2016 Puppet, Inc.
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


LICENSE
```diff
@@ -186,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright 2013 Puppet Labs
+   Copyright 2016 Puppet, Inc.
```
ferventcoder: Shouldn't this be 2013-2016?

ferventcoder:
Looks good besides my one comment. While you are at it you should add the maintainers section to the readme.

LICENSE
glennsarti: Hrmm...Don't think so. The license on Puppet doesn't/didn't have it.  Although I was the one that updated it.

glennsarti:
Probably.  Does that come in as part of modsync? if not, should it?

jtappa:
Looks good to me, the date range for copyright isn't necessary. Current year is standard for web. For instance, see below Github only lists current year.

glennsarti:
Looks like I targetted the wrong branch. Closing the PR

PR #99, opened 2016-05-06
(PA-285) Branding name change

jpogran:
Update all references from puppetlabs to puppet


ferventcoder:
Same question - better for stable?

ferventcoder:
Superseded by #101

PR #100, opened 2016-05-10
(MODULES-3344) Nano Server Compatibility

Iristyle:
- Use the CoreCLR compatible `WaitHandle.WaitOne(timeout)`
- Don't invoke `powershell.exe` with `-Sta`

### DIFF
--- a/lib/puppet/provider/exec/powershell.rb
+++ b/lib/puppet/provider/exec/powershell.rb
@@ -44,7 +44,7 @@ Puppet::Type.type(:exec).provide :powershell, :parent => Puppet::Provider::Exec 
   end
 
   def self.powershell_args
-    ps_args = ['-NoProfile', '-NonInteractive', '-Sta', '-NoLogo', '-ExecutionPolicy', 'Bypass', '-Command']
+    ps_args = ['-NoProfile', '-NonInteractive', '-NoLogo', '-ExecutionPolicy', 'Bypass', '-Command']
     ps_args << '-' if PuppetX::PowerShell::PowerShellManager.supported?
     ps_args
   end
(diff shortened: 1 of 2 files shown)


Iristyle:
Related to https://github.com/puppetlabs/puppetlabs-dsc/pull/199

Iristyle:
@puppetlabs/windows would like to get this one in prior to cutting our release.

jpogran:
While I'm fine with this on dsc, a concern I have with this in powershell is that there is a greater surface area here than with dsc. I know our unit/integration tests pass, but we don't do any COM interop or threading in these tests and STA operates differently than MTA in these cases. I will try to find examples for my concern, but if I don't find them by EOD I don't think this PR should wait for it and should merge without me.

Iristyle:
@jpogran we've never specified `-Sta` in prior versions of the module, so this PR is more consistent with v1, where it's unspecified.

The reality is that depending on which version of PowerShell you run on top of, you will get different behavior in our old module.

```
In Windows PowerShell 3.0, single-threaded apartment (STA) is the default. In Windows PowerShell 2.0, multi-threaded apartment (MTA) is the default.
```

Given it was previously a grab-bag based on PS version, I'm OK with it remaining that way, where Nano doesn't have a `-Sta`.

STA vs MTA is really about thread safety for the COM components themselves.  The tl; dr is that you should use STA when the COM objects you're consuming are not thread safe.  The old Chris Brumme post is a great one for covering this in detail / discussing how this impacts .NET code - https://blogs.msdn.microsoft.com/cbrumme/2004/02/02/apartments-and-pumping-in-the-clr/  (though as a console app, we shouldn't have to worry much about the message pump).

[Essential COM](https://books.google.com/books/about/Essential_COM.html?id=kfRWvKSePmAC) is also a good reference.

jpogran:
👍

PR #103, opened 2016-05-19
Update powershell.rb

wmuizelaar:
Make this module work on (old) ruby 1.8.7 as well.

### DIFF
--- a/lib/puppet/provider/exec/powershell.rb
+++ b/lib/puppet/provider/exec/powershell.rb
@@ -1,5 +1,5 @@
 require 'puppet/provider/exec'
-require_relative '../../../puppet_x/puppetlabs/powershell_manager'
+require 'puppet_x/puppetlabs/powershell_manager'
 
 Puppet::Type.type(:exec).provide :powershell, :parent => Puppet::Provider::Exec do
   confine :operatingsystem => :windows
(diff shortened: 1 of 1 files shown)


lib/puppet/provider/exec/powershell.rb
```diff
@@ -1,5 +1,5 @@
 require 'puppet/provider/exec'
-require_relative '../../../puppet_x/puppetlabs/powershell_manager'
+require 'puppet_x/puppetlabs/powershell_manager'
```
joshcooper: This will work for agents (because they add the pluginsync'ed lib directory to $LOAD_PATH). But it will cause catalog compilations to fail on the master. The master will `require` the powershell provider, but will be unable to require the `powershell_manager` helper code. This is why we prior to `require_relative` we had two different code paths for loading helper code, e.g. https://github.com/puppetlabs/puppetlabs-registry/blob/master/lib/puppet/provider/registry_key/registry.rb#L8-L11

Note if the agent on the master has done a pluginsync, then you won't see the problem. But if you delete `puppet agent --configprint libdir` on the master you will.

wmuizelaar: Ok. So, basically, to make sure we don't run into the issue that require_relative isn't available on ruby 1.8.7, I should alter this pull request to use the same way as has been done in the puppetlabs-registry module?

PR #105, opened 2016-05-23
(FM-5240) remove conflict between powershell & dsc

binford2k:
When the modules are installed in the same environment the
PowerShellManager class gets pluginsynced from both modules. The DSC one
wins always, which means that execs explode on linux nodes.


Iristyle:
Thanks for reporting - I think we're going to roll with a simpler approach for now.

PR #106, opened 2016-05-23
(FM-5240) Prevent powershell_manager DSC conflict

Iristyle:
- DSC and PowerShell both used the path of
   lib/puppet_x/puppetlabs/powershell_manager.b for their uniquely
   namespaced PowerShellManager class.

   Unfortunately, while the file was namespaced differently, upon
   extraction, the files would collide / overwrite one another, so
   only the last file extracted would win.

### DIFF
--- a/lib/puppet/provider/exec/powershell.rb
+++ b/lib/puppet/provider/exec/powershell.rb
@@ -1,5 +1,5 @@
 require 'puppet/provider/exec'
-require_relative '../../../puppet_x/puppetlabs/powershell_manager'
+require_relative '../../../puppet_x/puppetlabs/powershell/powershell_manager'
 
 Puppet::Type.type(:exec).provide :powershell, :parent => Puppet::Provider::Exec do
   confine :operatingsystem => :windows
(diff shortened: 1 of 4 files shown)


Iristyle:
Supercedes #105

lib/puppet/provider/exec/powershell.rb
```diff
@@ -1,5 +1,5 @@
 require 'puppet/provider/exec'
-require_relative '../../../puppet_x/puppetlabs/powershell_manager'
+require_relative '../../../puppet_x/puppetlabs/powershell/powershell_manager'
```
glennsarti: 👍  I like this naming standard better puppet_x/<company>/<module>/<class>

Iristyle: Should we change to `puppet_x/puppet/powershell/powershell_manager`?

glennsarti: Maybe.  I'm not bothered either way.

Iristyle: Yeah, let's not worry about it for now.

ferventcoder:
Other than "Why an FM issue?", I'm 👍

Iristyle:
Agreed on "why FM" @ferventcoder - I probably should have moved the ticket over first.

ferventcoder:
It can be moved now and it will redirect just fine. 👍

ferventcoder:
maybe? ;)

PR #110, opened 2016-05-25
Update powershell.rb

wmuizelaar:
Make this module work on (old) ruby 1.8.7 as well.

### DIFF
--- a/lib/puppet/provider/exec/powershell.rb
+++ b/lib/puppet/provider/exec/powershell.rb
@@ -1,5 +1,6 @@
 require 'puppet/provider/exec'
-require_relative '../../../puppet_x/puppetlabs/powershell/powershell_manager'
+module_base = Pathname.new(__FILE__).dirname
+require module_base + "../../../" + 'puppet_x/puppetlabs/powershell/powershell_manager'
 
 Puppet::Type.type(:exec).provide :powershell, :parent => Puppet::Provider::Exec do
   confine :operatingsystem => :windows


Iristyle:
@wmuizelaar we don't support Ruby 1.8.7 anymore (haven't for a long time actually).

I'm curious as to how you have an agent that's running on Ruby 1.8.7?  I assume this is a Linux node somewhere in your infrastructure that is failing to parse the module code?

dgarbus:
This is being pulled in as a dependency for one of our modules and breaking our Linux nodes as well:

```
Error: Could not autoload puppet/provider/exec/powershell: undefined method `require_relative' for main:Object
Error: Could not autoload puppet/type/exec: Could not autoload puppet/provider/exec/powershell: undefined method `require_relative' for main:Object
Error: Could not retrieve catalog from remote server: Could not intern from text/pson: Could not autoload puppet/type/exec: Could not autoload puppet/provider/exec/powershell: undefined method `require_relative' for main:Object
```

Iristyle:
@dgarbus I will ask the same question of you regarding Ruby 1.8.7 - I would be interested in knowing why such an old Ruby version is still running.  What version of Puppet is running on your master / Linux agents / Windows agents?

Thanks!

/cc @kenazk @ahpook

dgarbus:
We are running Puppet 3.8.X installed from the official PuppetLabs yum repo, which as far as I can tell, is still a supported version. For our OS, we are using CentOS 6.X, which uses ruby 1.8.7 as its system ruby. I can't imagine that we are the only ones bumping into this as CentOS 6 is still widely used, and the packages we are using are from PuppetLabs itself.

I only found this issue because we have another module that pulls in another module that pulls in this module as a dependency. To get around these errors for now, I simply locked our Puppetfile to pull in the latest 1.X release of this module.

wmuizelaar:
We are also running a lot of RHEL6 systems on puppet 3.8.7; and that version still runs on ruby 1.8.7.

Iristyle:
Ah, so @dgarbus and @wmuizelaar, you're both open source users - now I understand.

For our officially supported Enterprise products, we bundle Ruby and therefore do not support 1.8.7.  Further, Our Puppet 4 series has synced Ruby versions across our supported platforms for a number of reasons, including an effort to ease our support burden for issues like these, and to increase our ability to ship code more quickly.  Note that even the 3.x series is approaching end-of-life for our commercial users - https://groups.google.com/forum/#!topic/puppet-dev/jkcCNSk9w4w

Aside from the Puppet 3.x branch in Travis - https://github.com/puppetlabs/puppet/blob/3.x/.travis.yml we don't actively target / test against Ruby 1.8.7 for any module content that I'm aware of, so there may be breakages from time to time as a result of using that old Ruby 1.8.7.  Now might be a good time to think about upgrading your infrastructure to newer versions of Ruby / Puppet / PE to avoid any future downtime.

Since the fix here is pretty straightforward in this case, we will take care of updating, but I'm not sure yet of when a release may be kicked off.


Thanks for the report!

Iristyle:
Closing in favor of #114

PR #114, opened 2016-05-27
(maint) Remove require_relative for Ruby 1.8.7

Iristyle:
- Because this module can be run on Puppet 3.8, which can run on
   the system Ruby of a Linux machine (such as CentOS 6), it's
   possible that the cdoe will be parsed in a Ruby 1.8.7 environment.

   Move to require with a manually constructed path, which is the
   effectively the same thing as require_relative in newer Rubies.


Iristyle:
Supercedes #110

Iristyle:
@puppetlabs/windows this is ready for merge... please merge up to master once it's merged to stable here.

Iristyle:
Merged up to master in 418ffbd

PR #117, opened 2016-06-01
(MODULES-2634) Remove unnecessary binary test file

glennsarti:
A binary test file was committed in error in MODULES-2634.  This commit removes the test file and updates gitignore so that test files in that location (tmp/...) will be ignored in future


Iristyle:
👍

PR #118, opened 2016-06-02
(MODULES-2634) Fix tests for non-Windows agents

Iristyle:
- These tests were initially validated against a local definition
   that only included Windows agents.

   However, in CI, there are masters that are configured as agents as
   well, and some of these tests don't make sense.

   Refactor all the tests to run manifests only on Windows agents.


Iristyle:
Ran locally with a RHEL and Windows agent to ensure that tests are now confined to Windows agents and there are no more failures. Results:

```
Finished in 7 minutes 10 seconds (files took 5 minutes 36 seconds to load)
27 examples, 0 failures
```

Issue #123, opened 2016-07-05
cwd param dont work as expected


mightydok:
After upgrade to 2.0.1 version of plugin i notice that cwd param broken, and dont change path to value.
Example script:

```
  exec { "cwd test":
    cwd      => 'c:/windows/',
    command  => 'get-location',
    provider =>  powershell,
  }
```
At 1.0.6 version all works fine.

glennsarti:
@mightydok Thanks for the issue report.  I've reproduced the issue as well.  Would you be able to raise a ticket as per https://github.com/puppetlabs/puppetlabs-powershell/blob/master/CONTRIBUTING.md

glennsarti:
@mightydok To speed up resolution of this issue, I've raised https://tickets.puppetlabs.com/browse/MODULES-3565 on your behalf.

mightydok:
@glennsarti thnx for issue at puppetlabs.com. 
About CONTRIBUTING.md - i dont know what i need to do, this file describe how make PR. What should i do?

glennsarti:
@mightydok There is a section in the CONTRIBUTING section about how to raise an ticket.  However as I've done that for you already, there's nothing required of you.

mightydok:
@glennsarti oh, ok. Thnx again.

ferventcoder:
Since we have an issue linked up with this, I'm going to close this ticket.

PR #125, opened 2016-07-05
(MODULES-3565) Change the working directory if specified in the resource 

glennsarti:
Previously  the cwd resource parameter was not being used during powershell
script execution.  This PR checks to see if the cwd parameter is specified
on the resource and if so, inserts a 'Set-Location' command at the very top of
the powershell script to change the working directory prior to script execution.

The working directory will always default to the current directory of the process
which creates the powershell manager.  The working directory is not modified
across runspaces/

This PR also moved the default execution timeout value to a separate class
function so that the execute function can be invoked correctly i.e. If the
execute method is to be called with a custom working directory, but with a default
timeout.

### DIFF
--- a/lib/puppet/provider/exec/powershell.rb
+++ b/lib/puppet/provider/exec/powershell.rb
@@ -68,7 +68,12 @@ Puppet::Type.type(:exec).provide :powershell, :parent => Puppet::Provider::Exec 
         return super("cmd.exe /c \"\"#{native_path(command(:powershell))}\" #{legacy_args} -Command - < \"#{native_path}\"\"", check)
       end
     else
-      result = ps_manager.execute(command)
+      working_dir = resource[:cwd]
+      if (!working_dir.nil?)
+        self.fail "Working directory '#{working_dir}' does not exist" unless File.directory?(working_dir)
+      end
+
+      result = ps_manager.execute(command,nil,working_dir)
 
       stdout      = result[:stdout]
       stderr      = result[:stderr]
--- a/lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
+++ b/lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
@@ -41,11 +41,11 @@ module PuppetX
         at_exit { exit }
       end
 
-      def execute(powershell_code, timeout_ms = 300 * 1000)
+      def execute(powershell_code, timeout_ms = nil, working_dir = nil)
         output_ready_event_name =  "Global\\#{SecureRandom.uuid}"
         output_ready_event = self.class.create_event(output_ready_event_name)
 
-        code = make_ps_code(powershell_code, output_ready_event_name, timeout_ms)
+        code = make_ps_code(powershell_code, output_ready_event_name, timeout_ms, working_dir)
 
         out, err = exec_read_result(code, output_ready_event)
 
@@ -104,7 +104,14 @@ module PuppetX
         template.result(binding)
       end
 
-      def make_ps_code(powershell_code, output_ready_event_name, timeout_ms = 300 * 1000)
+      def make_ps_code(powershell_code, output_ready_event_name, timeout_ms = nil, working_dir = nil)
+        begin
+          timeout_ms = Integer(timeout_ms)
+          # Lower bound protection. The polling resolution is only 50ms
+          if (timeout_ms < 50) then timeout_ms = 50 end
+        rescue
+          timeout_ms = 300 * 1000
+        end
         <<-CODE
 $params = @{
   Code = @'
@@ -112,6 +119,7 @@ $params = @{
 '@
   EventName = "#{output_ready_event_name}"
   TimeoutMilliseconds = #{timeout_ms}
+  WorkingDirectory = "#{working_dir}"
 }
 
 Invoke-PowerShellUserCode @params
--- a/lib/puppet_x/templates/init_ps.ps1.erb
+++ b/lib/puppet_x/templates/init_ps.ps1.erb
@@ -294,6 +294,7 @@ function New-XmlResult
 }
 
 Add-Type -TypeDefinition $hostSource -Language CSharp
+$global:DefaultWorkingDirectory = (Get-Location -PSProvider FileSystem).Path
 
 #this is a string so we can import into our dynamic PS instance
 $global:ourFunctions = @'
@@ -359,7 +360,10 @@ function Invoke-PowerShellUserCode
     $EventName,
 
     [Int]
-    $TimeoutMilliseconds
+    $TimeoutMilliseconds,
+
+    [String]
+    $WorkingDirectory    
   )
 
 
@@ -392,6 +396,12 @@ function Invoke-PowerShellUserCode
     [Void]$ps.AddScript($global:ourFunctions)
     $ps.Invoke()
 
+    if ([string]::IsNullOrEmpty($WorkingDirectory)) {
+      $ps.Runspace.SessionStateProxy.Path.SetLocation($global:DefaultWorkingDirectory)    
+    } else {
+      if (-not (Test-Path -Path $WorkingDirectory)) { Throw "Working directory `"$WorkingDirectory`" does not exist" }
+      $ps.Runspace.SessionStateProxy.Path.SetLocation($WorkingDirectory)
+    }
 
     if(!$global:environmentVariables){
       $ps.Commands.Clear()
--- a/spec/unit/provider/exec/powershell_spec.rb
+++ b/spec/unit/provider/exec/powershell_spec.rb
@@ -108,6 +108,25 @@ describe Puppet::Type.type(:exec).provider(:powershell) do
     end
   end
 
+  describe 'when specifying a working directory' do
+    describe 'that does not exist' do
+      let(:work_dir)  {
+        if Puppet.features.microsoft_windows?
+          "#{ENV['SYSTEMROOT']}\\some\\directory\\that\\does\\not\\exist"
+        else
+          '/some/directory/that/does/not/exist'
+        end
+      }
+      let(:command)  { 'exit 0' }
+      let(:resource) { Puppet::Type.type(:exec).new(:command => command, :provider => :powershell, :cwd => work_dir) }
+      let(:provider) { described_class.new(resource) }
+
+      it 'emits an error when working directory does not exist' do
+        expect { provider.run(command) }.to raise_error(/Working directory .+ does not exist/) 
+      end
+    end
+  end
+
   describe 'when applying a catalog' do
     let(:manifest) { <<-MANIFEST
       exec { 'PS':
--- a/spec/integration/puppet_x/puppetlabs/powershell_manager_spec.rb
+++ b/spec/integration/puppet_x/puppetlabs/powershell_manager_spec.rb
@@ -307,7 +307,7 @@ $bytes_in_k = (1024 * 64) + 1
     it "should handle shell redirection" do
       # the test here is to ensure that this doesn't break. because we merge the streams regardless
       # the opposite of this test shows the same thing
-      result = manager.execute('function test-warning{ ps;write-warning \'foo\' }; test-warning 3>&1')
+      result = manager.execute('function test-error{ ps;write-error \'foo\' }; test-error 2>&1')
 
       expect(result[:stdout]).not_to eq(nil)
       expect(result[:exitcode]).to eq(0)
--- a/spec/integration/puppet_x/puppetlabs/powershell_manager_spec.rb
+++ b/spec/integration/puppet_x/puppetlabs/powershell_manager_spec.rb
@@ -204,6 +204,43 @@ $bytes_in_k = (1024 * 64) + 1
 
       expect(result[:errormessage]).not_to be_empty
     end
+
+    it "should error if working directory does not exist" do
+      work_dir = 'C:/some/directory/that/does/not/exist'
+
+      result = manager.execute('(Get-Location).Path',nil,work_dir)
+
+      expect(result[:exitcode]).to_not eq(0)
+      expect(result[:errormessage]).to match(/Working directory .+ does not exist/)
+    end
+
+    it "should allow forward slashes in working directory" do
+      work_dir = ENV["WINDIR"]
+      forward_work_dir = work_dir.gsub('\\','/')
+
+      result = manager.execute('(Get-Location).Path',nil,work_dir)[:stdout]
+
+      expect(result).to eq("#{work_dir}\r\n")
+    end
+
+    it "should use a specific working directory if set" do
+      work_dir = ENV["WINDIR"]
+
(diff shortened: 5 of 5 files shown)


glennsarti:
Relates to https://github.com/puppetlabs/puppetlabs-powershell/issues/123

lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
```diff
@@ -104,10 +108,21 @@ def make_ps_init_code(init_ready_event_name)
...
+          if Puppet::Util::Platform.windows?
+            working_dir = working_dir.gsub(File::SEPARATOR, File::ALT_SEPARATOR)
+          end
+          set_ps_location = "Set-Location \"#{working_dir}\" | Out-Null\n\r" unless working_dir == nil
+        end
+
         <<-CODE
 $params = @{
   Code = @'
+#{set_ps_location}
```
jpogran: Did you try multi-line commands (something more complicated than a single cmdlet) with this? Afraid a single newline won't be enough of a seperation here.

Also, what happens if set-location fails? Shouldn't this be two executions?

lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
```diff
@@ -104,10 +108,21 @@ def make_ps_init_code(init_ready_event_name)
...
       end
 
-      def make_ps_code(powershell_code, output_ready_event_name, timeout_ms = 300 * 1000)
+      def make_ps_code(powershell_code, output_ready_event_name, timeout_ms = default_execute_timeout, working_dir = nil)
+        # Convert the working directory to a native pathname, if required
+        if (working_dir == nil)
+          set_ps_location = ''
+        else
+          if Puppet::Util::Platform.windows?
+            working_dir = working_dir.gsub(File::SEPARATOR, File::ALT_SEPARATOR)
```
jpogran: Is changing path separators really a good idea here? UNC paths or volume paths come to mind.

glennsarti: Old behaviour in 1.0.6 uses Puppet code;
https://github.com/puppetlabs/puppet/blob/master/lib/puppet/provider/exec.rb#L14-L29

lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
```diff
@@ -104,10 +108,21 @@ def make_ps_init_code(init_ready_event_name)
...
+          end
+          set_ps_location = "Set-Location \"#{working_dir}\" | Out-Null\n\r" unless working_dir == nil
```
jpogran: Nit - use `-Path` parameter instead of default parameter resolution

lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
glennsarti: Forward and backslashes are not valid characters in directory names so it is safe to convert.  This step was to ensure that we give PowerShell windows native looking paths instead of the Puppet style forward slash based ones.

lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
glennsarti: I'm not sure how else to separate the two pieces of code.  A single line feed should be enough for PS to know the end of the `Set-Location....` and beginning of any other proceeding text.

Not sure about the failure.  Should that be terminating? probably.  Chances of it happening? Low.  There is an existance check in the provider so whatever path is passed in must exist.  Although permissions could be an issue i.e. user has rights to read that the directory exists, but not traversal.

lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
glennsarti: Noted. Will fix

lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
```diff
@@ -9,6 +9,8 @@ module PowerShell
     class PowerShellManager
       extend FFI::Library if Puppet::Util::Platform.windows?
 
+      attr_reader :default_execute_timeout
```
Iristyle: Is this from another PR / commit?

lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
Iristyle: I think we normally use `expand_path` to canonicalize paths in this case.

But PS handles mixed slashes correctly anyhow, no?

spec/unit/provider/exec/powershell_spec.rb
```diff
@@ -107,6 +107,32 @@
...
+      let(:provider) { described_class.new(resource) }
+
+      it 'sets the working directory in the poweshell session' do
+        output, status = provider.run(command)
+        expect(output).to eq("#{work_dir}\r\n")
+        expect(status.exitstatus).to eq(0)
+      end
+    end
+  end
+
```
Iristyle: What about a test ensuring that CWD doesn't carry over from one run to another?

lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
Iristyle: Rather than `Set-Location`, I generally prefer to do a `Push-Location` / `Pop-Location` pair to reset back to the state before changing the directory.

Iristyle: James has noted that maybe it's best to use the runspace - http://stackoverflow.com/questions/896306/setting-the-start-dir-when-calling-powershell-from-net

Then we could more easily save / restore the settings prior to each run (like we do for process level env vars)

spec/unit/provider/exec/powershell_spec.rb
glennsarti: Updated.

lib/puppet/provider/exec/powershell.rb
```diff
@@ -68,7 +68,13 @@ def run(command, check = false)
         return super("cmd.exe /c \"\"#{native_path(command(:powershell))}\" #{legacy_args} -Command - < \"#{native_path}\"\"", check)
       end
     else
-      result = ps_manager.execute(command)
+      if working_dir = resource[:cwd]
+        self.fail "Working directory '#{working_dir}' does not exist" unless File.directory?(working_dir)
```
ferventcoder: When will this be hit?

lib/puppet/provider/exec/powershell.rb
```diff
@@ -68,7 +68,13 @@ def run(command, check = false)
         return super("cmd.exe /c \"\"#{native_path(command(:powershell))}\" #{legacy_args} -Command - < \"#{native_path}\"\"", check)
       end
     else
-      result = ps_manager.execute(command)
+      if working_dir = resource[:cwd]
```
ferventcoder: I've never been a huge fan of an if that is also a setter.

lib/puppet/provider/exec/powershell.rb
glennsarti: This is hit when a resource is executed (onlyif, unless or command).  It's a guard to stop trying to execute Powershell if the working dir doesn't exist.  There is a guard in the powershell manager similar to this.  In practice this particular check isn't required as it will be picked up in the PS manager but I preferred to fail early.

Thinking about it, this will fail the resource in and onlyif/unless situation whereas If I just left it in the PS manager, it will evaluate to a non-zero exit code.  These scenarios are different

lib/puppet/provider/exec/powershell.rb
glennsarti: Noted.  Will fix.

lib/puppet/provider/exec/powershell.rb
ferventcoder: I was trying to understand mostly if this is a good thing or not. Is `{working_dir}` always and only set explicitly by the user?

glennsarti: If using a resource/manifest yes.

lib/puppet_x/templates/init_ps.ps1.erb
```diff
@@ -391,6 +395,12 @@ function Reset-ProcessPowerShellVariables
     [Void]$ps.AddScript($ourFunctions)
     $ps.Invoke()
 
+    if ($WorkingDirectory -ne '') {
```
ferventcoder: I always tend to overguard on empty strings - `if ($WorkingDirectory -ne $null -and $WorkingDirectory -ne '')`

ferventcoder: This may be more than is necessary.

lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
glennsarti: I needed to add an additional parameter to the execute function.  However I needed to give a value as the parameter is not optional.  I needed to expose the default timeout on the powershell manager object so that I could use it in `lib/puppet/provider/exec/powershell.rb`.  So I created a readonly property.  Not sure if this should be in a different commit or PR.  Also note I needed this value in the integration tests too.

lib/puppet_x/templates/init_ps.ps1.erb
glennsarti: The $WorkingDirectory in this case is always cast to a string due to Line 119 in powershell_manager.rb
```
...
WorkingDirectory = "#{working_dir}"
...
```

lib/puppet_x/templates/init_ps.ps1.erb
```diff
@@ -391,6 +395,12 @@ function Reset-ProcessPowerShellVariables
...
+      if (-not (Test-Path -Path $WorkingDirectory)) { Throw "Working directory `"$WorkingDirectory`" does not exist" }
+      $ps.Runspace.SessionStateProxy.Path.SetLocation($WorkingDirectory)
+    } else {
+      $ps.Runspace.SessionStateProxy.Path.SetLocation($global:DefaultWorkingDirectory)
```
jpogran: Nit: Extra spaces here

lib/puppet_x/templates/init_ps.ps1.erb
jpogran: Would have used `[string]::IsNullOrEmpty()` myself

lib/puppet_x/templates/init_ps.ps1.erb
```diff
@@ -294,6 +294,7 @@ function New-XmlResult
 }
 
 Add-Type -TypeDefinition $hostSource -Language CSharp
+$global:DefaultWorkingDirectory = (Get-Location).Path
```
jpogran: This will produce bugs with difference between .NET's idea of cwd and PowerShell's. Can't remember link off top of head, but we have to specify filesystem provider. Looking now

lib/puppet_x/templates/init_ps.ps1.erb
glennsarti: I'm not seeing where the extra spaces are.  The `if (-not ...` is in the right spot

lib/puppet_x/templates/init_ps.ps1.erb
glennsarti: Modified it to use the location of the FileSystem Provider

lib/puppet_x/templates/init_ps.ps1.erb
glennsarti: Modified to use IsNullOrEmpty

lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
```diff
@@ -41,11 +41,11 @@ def initialize(cmd)
         at_exit { exit }
       end
 
-      def execute(powershell_code, timeout_ms = 300 * 1000)
+      def execute(powershell_code, timeout_ms = nil, working_dir = nil)
```
MosesMendoza: I'm trying to figure out how `timeout_ms` gets set because it seems to me it will always be `nil`, and the only place I see it getting used is in the above invocation  `ps_manager.execute` at powershell.rb:76. I can't seem to find it getting passed in from the resource anywhere via `resource[:timeout]`.  I'm sure I've confused something - is `timeout_ms` something different from the `timeout` resource parameter?

glennsarti: 1. `timeout_ms` ultimately gets used in the `make_ps_code` function where it sets the default for 300*1000.
2. resource[:timeout] incidentally is set to `300.0` if you don't specify anything in the manifest
3. This is protection for when the method is called directly (mainly tests).  Normally this is only called via the provider

MosesMendoza: Oh I see. Thanks. Somehow I had completely missed https://github.com/puppetlabs/puppetlabs-powershell/pull/136 and MODULES-3709 and was confused why the timeout resource param didn't appear to be connected anywhere. Got it now.

MosesMendoza:
👍 . Also did a couple basic smoke tests as captured in https://gist.github.com/MosesMendoza/0fcb70e884be8dbab3979a293a762b03.

lib/puppet_x/templates/init_ps.ps1.erb
```diff
@@ -392,6 +396,12 @@ function Invoke-PowerShellUserCode
     [Void]$ps.AddScript($global:ourFunctions)
     $ps.Invoke()
 
+    if ([string]::IsNullOrEmpty($WorkingDirectory)) {
+      $ps.Runspace.SessionStateProxy.Path.SetLocation($global:DefaultWorkingDirectory)    
+    } else {
+      if (-not (Test-Path -Path $WorkingDirectory)) { Throw "Working directory `"$WorkingDirectory`" does not exist" }
+      $ps.Runspace.SessionStateProxy.Path.SetLocation($WorkingDirectory)
```
MosesMendoza: One question - is there any situation where we wouldn't be able to `Se-tLocation` due to permissions but it would still be a valid path according to `Test-Path` ?  E.g. akin to on *nix without execute bit set? My quick testing showed no - always able to enter the most locked down folders, but I'm not sure.

glennsarti: Looks like this is ok

So a quick test:

* Created `C:\a\b\c`
* Set Deny Everyone All perms on `C:\a\b` only (This Folder and files only)

* Can `Set-Location` to `C:\a`,  `C:\a\b` and  `C:\a\b\c`
* Can only do `Get-ChildItem`  on  `C:\a` and  `C:\a\b\c`.
* `Get Access Denied` on`C:\a\b` 
* Can `Test-Path` to `C:\a`,  `C:\a\b` and  `C:\a\b\c` all with `$true`

Side note:
CMD.exe does have that problem.  Can't `cd c:\a\b\` - get's access denied.

PR #127, opened 2016-07-08
(MODULES-3468) Update README.md

glennsarti:
This commit
- Adds the minimum powershell version to the Limitations section
- Removes Window 2003 as a supported OS for this module
- Adds a space in the markdown headers between the hash and text.  While this
renders correctly in GitHub and the Puppet Forge some editors and parsers
require a white space.

[ci skip]

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
-#powershell
+# powershell
 
-####Table of Contents
+#### Table of Contents
 
 1. [Overview](#overview)
 2. [Module Description - What the module does and why it is useful](#module-description)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,6 +1,6 @@
-#powershell
+# PowerShell
```
ferventcoder: This is typically lowercase, but you will need to ask @jbondpdx why as a refresher. I believe it is to show this is the provider/module versus the underlying technology.

Iristyle:
Looks good, but I don't think we need the `(doc)` in the commit subject...

README.md
Iristyle: 👍 thanks @ferventcoder

glennsarti: Fixed

glennsarti:
Commit modified without the `doc`

README.md
jbondpdx: Yep, that's correct @ferventcoder . Module names (powershell) are lowercased, the technology names are capitalized as per their usual style (PowerShell).

PR #129, opened 2016-07-12
(MODULES-3399) Exit user scripts via $LASTEXITCODE 

Iristyle:
- Previously in v1 of the module, if the call operator & was used to
   execute an external script, and that script performed an "exit 1",
   that exit code would propagate.  This technique could be used within
   an unless or onlyif to ensure idempotent behavior like:

   unless => '& exit-one.ps1'

   This would allow users to create fully reusable PowerShell scripts
   that could be invoked outside of the Puppet environment during
   development and testing.

 - Since code is invoked differently in the v2 module, that behavior
   changed, and any exit codes set externally in a PS script invoked by
   the call operator are not propagating. The custom PowerShell host
   SetShouldExit method is not being triggered in this situation.

   This commit restores the old behavior by performing an
   exit $LASTEXITCODE after executing the user code.

   NOTE: this has the side effect of returning an exit code that may be
   unexpected, based on the last external command invoked by PowerShell.

   Users will be expected to understand this when developing longer more
   complex scripts.


spec/integration/puppet_x/puppetlabs/powershell_manager_spec.rb
```diff
@@ -47,6 +47,22 @@ class PowerShellManager; end
       expect(result[:exitcode]).to eq(1)
     end
 
+    it "should return the exitcode of the last command to set an exit code" do
+      result = manager.execute("$LASTEXITCODE = 0; write-output 'foo'; cmd.exe /c 'exit 99'; write-output 'bar'")
+
+      # STDERR is interpolating the newlines thus it's \n instead of the usual Windows \r\n
+      expect(result[:stdout]).to eq("foo\r\nbar\r\n")
+      expect(result[:exitcode]).to eq(99)
```
Iristyle: This might be an unexpected result now, but arguably it is "correct".

Worth discussing @puppetlabs/windows

glennsarti: Seems correct to me.  LASTEXITCODE should return the last exitcode emitted by an external entity.

ferventcoder: I know tools like `reg.exe` (aka `reg add`, `reg query`, etc) will write to stderr instead of stdout. I believe they exit correctly though. Is there a place for interpretation where this could go bad?

I'm believing no. 👍

glennsarti: Java also does this (java -version).  Correct exitcode but writes to STDERR

lib/puppet_x/templates/init_ps.ps1.erb
```diff
@@ -424,6 +425,10 @@ function Reset-ProcessPowerShellVariables
     # and writes it to the PSHost we create
     # this needs to be the last thing executed
     [void]$ps.AddCommand("out-default");
+
+    # if the call operator & establishedd an exit code, exit with it
```
ferventcoder: micro - misspelling

MosesMendoza:
👍 . Pulled this is in and validated against a trivial manifest as supplied in MODULES-3399

ferventcoder:
Validating this against a more heavy weight script currently.

ferventcoder:
Looks good to me 👍

MosesMendoza:
one thing that seems a bit odd is I'm not sure `returns` param is working quite right:

```
C:\Users\moses\development\puppet(stable)> type C:\fail.ps1
exit 42
C:\Users\moses\development\puppet(stable)> type C:\Users\moses\development\tester_class\site.pp
exec { 'should return 42':
  command   => '& C:\fail.ps1',
  provider  => 'powershell',
  returns   => 42,
}
C:\Users\moses\development\puppet(stable)> be puppet apply C:\Users\moses\development\tester_class\site.pp
DL is deprecated, please use Fiddle
Notice: Compiled catalog for win-ghb4gdkiluo.localdomain in environment production in 0.16 seconds
Error: & C:\fail.ps1 returned 1 instead of one of [42]
Error: /Stage[main]/Main/Exec[should return 42]/returns: change from notrun to 42 failed: & C:\fail.ps1 returned 1 instead of one of [42]
Notice: Applied catalog in 1.08 seconds
```

PR #135, opened 2016-08-23
(maint) Fix test failures for Puppet 4.6

Iristyle:
- Some internal changes to transaction behavior that are part of
   Puppet 4.6, have induced a new problem with the existing tests that
   invoked the compiler.  Transactionstore.yaml files are now
   produced when a transaction is processed, for the sake of
   identifying changes as part of the `corrective changes` feature.

   Without the Puppet[:statedir] explicitly set, compiling a catalog
   and executing a transaction will attempt to write a YAML file to
   c:\dev\null.

   Instead, explicitly set the path to allow for a real file to be
   written as part of the transaction.

### DIFF
--- a/spec/unit/provider/exec/powershell_spec.rb
+++ b/spec/unit/provider/exec/powershell_spec.rb
@@ -2,6 +2,7 @@
 require 'spec_helper'
 require 'puppet/util'
 require 'puppet_x/puppetlabs/powershell/powershell_manager'
+require 'fileutils'
 
 describe Puppet::Type.type(:exec).provider(:powershell) do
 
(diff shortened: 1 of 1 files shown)


spec/unit/provider/exec/powershell_spec.rb
```diff
@@ -115,6 +116,19 @@
...
+      # a statedir setting must now exist per the new transactionstore code
+      # introduced in Puppet 4.6 for corrective changes, as a new YAML file
+      # called transactionstore.yaml will be written under this path
+      # which defaults to c:\dev\null when not set on Windows
+      Puppet[:statedir] = tmpdir
+    end
+
+    after :each do
+      FileUtils.rm_rf(tmpdir)
+    end
```
ferventcoder: This seems better to change as part of the spec helper (and then later as puppetlabs_spec_helper).

ferventcoder: Like somewhere in https://github.com/puppetlabs/puppetlabs-powershell/blob/f8ef8bbd4bcd34602187fdcc6cceec763c2abc40/spec/spec_helper.rb#L24

ferventcoder: No worries, this one is simpler as there is one place to adjust. I was just thinking for consistency if we add it in several places when we get repos that have more than one test file ;)

Iristyle: Yeah, I could have moved it further out... but... there's also the matter of this only being applicable to a couple of tests.  And those tests are a bit outside the norm of our regular unit tests given they use Puppet as a library to compile manifests.

Iristyle: I could also change this to `vardir` from `statedir`

glennsarti:
+1 for merge from me.  @ferventcoder you good for merge?

spec/unit/provider/exec/powershell_spec.rb
Iristyle: On second thought, we probably want `/dev/null` behavior for the sake of specs.

PSH should probably be updated to mock out the transaction store code, which most implementers will not need for the sake of writing tests like this / against `puppet apply`

PR #136, opened 2016-08-24
(MODULES-3709) Respect resource timeout interval in powershell manager

glennsarti:
Previously, the a timeout value could be specified in a resource however that
value was ignore and the default of 300s was used by the powershell manager.
This commit passes throught the timeout value from the resource into the
powershell manager and also adds an acceptance test to verify this behaviour.

This builds in PR #125

### DIFF
--- a/lib/puppet/provider/exec/powershell.rb
+++ b/lib/puppet/provider/exec/powershell.rb
@@ -68,7 +68,12 @@ Puppet::Type.type(:exec).provide :powershell, :parent => Puppet::Provider::Exec 
         return super("cmd.exe /c \"\"#{native_path(command(:powershell))}\" #{legacy_args} -Command - < \"#{native_path}\"\"", check)
       end
     else
-      result = ps_manager.execute(command)
+      working_dir = resource[:cwd]
+      if (!working_dir.nil?)
+        self.fail "Working directory '#{working_dir}' does not exist" unless File.directory?(working_dir)
+      end
+
+      result = ps_manager.execute(command,nil,working_dir)
 
       stdout      = result[:stdout]
       stderr      = result[:stderr]
--- a/lib/puppet/provider/exec/powershell.rb
+++ b/lib/puppet/provider/exec/powershell.rb
@@ -72,8 +72,9 @@ Puppet::Type.type(:exec).provide :powershell, :parent => Puppet::Provider::Exec 
       if (!working_dir.nil?)
         self.fail "Working directory '#{working_dir}' does not exist" unless File.directory?(working_dir)
       end
+      timeout_ms = resource[:timeout].nil? ? nil : resource[:timeout] * 1000
 
-      result = ps_manager.execute(command,nil,working_dir)
+      result = ps_manager.execute(command,timeout_ms,working_dir)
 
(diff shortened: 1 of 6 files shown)


lib/puppet/provider/exec/powershell.rb
```diff
@@ -68,7 +68,8 @@ def run(command, check = false)
         return super("cmd.exe /c \"\"#{native_path(command(:powershell))}\" #{legacy_args} -Command - < \"#{native_path}\"\"", check)
       end
     else
-      result = ps_manager.execute(command)
+      timeout_ms = resource[:timeout].nil? ? nil : resource[:timeout].to_i * 1000
```
Iristyle: `to_i` .. hmm... does this vary by Puppet version?

glennsarti: The type? I imagine it may come in as a string or float.

MosesMendoza: The exec type casts as a float so i'm not sure `to_i` is needed? (https://github.com/puppetlabs/puppet/blob/master/lib/puppet/type/exec.rb#L283-L290) We may lose some precision actually if we do.

spec/acceptance/exec_powershell_spec.rb
```diff
@@ -322,6 +322,17 @@
...
 
+  describe 'should error if timeout is exceeded' do
+    ptimeoutexception = <<-MANIFEST
+      exec{'PowershellException':
+        command  => 'Write-Host "Going to sleep now..."; Start-Sleep 5',
+        timeout  => 2,
+        provider => powershell,
+      }
+    MANIFEST
+    it_should_behave_like 'should fail', ptimeoutexception
```
MosesMendoza: It doesn't seem like there's anyway to catch an error code here to validate that it failed due to a timeout, so is it worth matching against something like `timeout.*exceeded` to ensure this failed due to timeout? I suppose we have been going away from string matching...

glennsarti: 1. Using this shared example, not I can't detect that
2. Yes you're right, the timeout errors always manifests as an Error.  Unlike https://github.com/puppetlabs/puppet/blob/master/lib/puppet/provider/exec.rb#L55 where it raises a `Timeout::Error` type

lib/puppet/provider/exec/powershell.rb
glennsarti: Powershell ultimately needs an integer.  Althought I believe you're right about precision.  Perhaps;

```
(resource[:timeout] * 1000).to_i
```

glennsarti: That being said....if https://github.com/puppetlabs/puppetlabs-powershell/pull/125 is merged prior, it does the integer conversion for us so the `to_i` is redundant.

MosesMendoza: Indeed! good catch. Then all that's needed here is just `resource[:timeout] * 1000`

glennsarti:
Given that this seems to touch the same code paths as PR #125, I'll rebase onto that

PR #141, opened 2016-09-06
(MODULES-3839) Emit better user code parsing exceptions

glennsarti:
Previously in MODULES-3443 the exception handler was updated to display more
meaningful errors however one type of error was not handled correctly.  If
Powershell generated an `IncompleteParseException` error, the message displayed
to the user would suggest that it occured in `throw $_.Exception.InnerException`
in a line and offset that probably didn't exist.  This is because unlike the
`ParseException` type which gives useful information, an
`IncompleteParseException` can not determine where in the script the error
occured.

This commit changes catches the `IncompleteParseException` error an rethrows it
as a simple text error.  This suppresses the `line:xxx char:xx` text from
being shown to the user, while still displaying the root cause of the error.

This commit refactors the tests slightly so that different error types are
exercised; Runtime, ParseException and IncompleteParseException.

### DIFF
--- a/lib/puppet_x/templates/init_ps.ps1.erb
+++ b/lib/puppet_x/templates/init_ps.ps1.erb
@@ -453,6 +453,9 @@ function Invoke-PowerShellUserCode
     try
     {
       $ps.EndInvoke($asyncResult)
+    } catch [System.Management.Automation.IncompleteParseException] {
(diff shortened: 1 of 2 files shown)


Iristyle:
Looks good on a quick pass

PR #142, opened 2016-09-07
(MODULES-3690) Use custom binary pipe IPC

Iristyle:
- Instead of using the stdin present by popen3, create a separate
   NamedPipe listener directly inside the PowerShell code.  This can
   be written to from Ruby by calling File.Open with a Windows style
   .\\.\pipe\ path to the pipe.  This File is a Ruby IO instance that
   behaves just like the existing streams returned by popen3.

   Send commands through this mechanism for several reasons:

   - Guarantee the protocol between Ruby client and PowerShell host,
     so that any normal PowerShell stdin processing problems are
     averted.  WMF 5.1 (Windows 10 anniversary update / Windows Server
     2016) accidentally made some breaking changes to the stdin
     parsing which causes existing code to completely fail.  This new
     implementation works around that problem completely.

   - Strictly control the encoding so that passing UTF-8 values is
     not a problem.  By using binary data in the pipe, and not
     converting to a string until the data has been fully read, this
     ensures there are no problems buffering the pipe data / treating
     it as a string.

     Add some basic passing UTF-8 tests.

   - All values passed to PowerShell from Ruby are length prefixed
     in number of bytes (of the current encoding, UTF-8) and lengths
     are passed as 32-bit Little Endian.

 - The PowerShell code now supports an encoding being passed in, in
   addition to the name of the named pipe that is being used. The
   encoding defaults to UTF-8 and the named pipe name must be passed
   in via Ruby.

   Some minor changes have been made to accomodate the new workflow

 - The manager, upon creation, must know whether it is operating in
   debug mode, so that it can pass that parameter through to the
   PowerShell bootstrap process, so that option has been added to
   its initializer.

 - While some commits have redirected output written over
   [Console]::Out / [Console]::Error to internal StringWriter
   instances in the custom PowerShell host, the possibility still
   exists that a native binary may access the underlying stdout /
   stderr streams, subverting the .NET driven redirection.

   While that remains unlikely, ensure that those streams continue to
   be drained, so that a deadlock never occurs. This output is
   captured and returned so that the provider may emit it if a user
   asks for it (via --debug or logoutput parameter).

 - If the number of bytes written to the pipe are less than expected
   throw an error indicating a broken pipe.

   The EPIPE thrown will be caught in exec_read_result like the
   existing pipe breakages.

   Some tests have been written to demonstrate this behavior.

- Length prefix binary responses back to Ruby from PowerShell so that
   a blocking read can be used within Ruby to wait on response.

   This removes the need to signal an event from PowerShell that Ruby
   waits on before it can stop draining pipes.  That previous
   implementation was very Win32 specific, which is a problem for
   making this module cross-platform.

 - Make an effort to always wait on completion of all threads inside
   the read_streams method.  Even in the face of crashes / unexpected
   behavior when reading the pipe, stdout or stderr, all reader
   threads are expected to cleanly exit.

 - Mark testing pending that seems to be triggering a w32_reset_event
   related segfault in Ruby when thread teardown occurs.  It's unknown
   why this crash it triggered by closing the stdout handle, but not
   the stderr handle.  The crash occurs intermittently.

 - Previously, an event name was passed to PowerShell, and it signaled
   the event when the pipe listener was ready.

   A simple File.Open style loop can instead be performed against the
   pipe path to poll it until it's ready.

   This removes a host of Win32 specific code, and the usage of FFI,
   which would prevent a cross-platform implementation.

   Adjustments will still have to be made to the code to make it
   fully cross-platform, but this code removal is a good start.

 - With a binary pipe, there is no reason to use XML anymore as
   newline terminators are irrelevant. Instead of calling a method
   that reads until an EOL terminator, length prefix the response,
   then send an arbitrary number of key value pairs in the format:

   - length of name (in UTF-8 bytes)
   - name
   - length of value (in UTF-8 bytes)
   - value

 - PowerShell no longer needs to use literal XML when building a
   response.

 - Ruby no longer needs to do anything with REXML

### DIFF
--- a/lib/puppet/provider/exec/powershell.rb
+++ b/lib/puppet/provider/exec/powershell.rb
@@ -44,13 +44,16 @@ Puppet::Type.type(:exec).provide :powershell, :parent => Puppet::Provider::Exec 
   end
 
   def self.powershell_args
-    ps_args = ['-NoProfile', '-NonInteractive', '-NoLogo', '-ExecutionPolicy', 'Bypass', '-Command']
-    ps_args << '-' if PuppetX::PowerShell::PowerShellManager.supported?
+    ps_args = ['-NoProfile', '-NonInteractive', '-NoLogo', '-ExecutionPolicy', 'Bypass']
+    ps_args << '-Command' if !PuppetX::PowerShell::PowerShellManager.supported?
+
     ps_args
   end
 
   def ps_manager
-    PuppetX::PowerShell::PowerShellManager.instance("#{command(:powershell)} #{self.class.powershell_args.join(' ')}")
+    debug_output = Puppet::Util::Log.level == :debug
+    manager_args = "#{command(:powershell)} #{self.class.powershell_args().join(' ')}"
+    PuppetX::PowerShell::PowerShellManager.instance(manager_args, debug_output)
   end
 
   def run(command, check = false)
@@ -77,6 +80,7 @@ Puppet::Type.type(:exec).provide :powershell, :parent => Puppet::Provider::Exec 
       result = ps_manager.execute(command,timeout_ms,working_dir)
 
       stdout      = result[:stdout]
+      native_out  = result[:native_stdout]
       stderr      = result[:stderr]
       exit_code   = result[:exitcode]
 
@@ -86,7 +90,7 @@ Puppet::Type.type(:exec).provide :powershell, :parent => Puppet::Provider::Exec 
 
       Puppet.debug "STDERR: #{result[:errormessage]}" unless result[:errormessage].nil?
 
-      output = Puppet::Util::Execution::ProcessOutput.new(stdout.to_s || '', exit_code)
+      output = Puppet::Util::Execution::ProcessOutput.new(stdout.to_s + native_out.to_s, exit_code)
 
       return output, output
     end
--- a/lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
+++ b/lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
@@ -87,29 +87,14 @@ module PuppetX
         # an error was caught during execution that has invalidated any results
         return { :exitcode => -1, :stderr => err } if !@usable && out.nil?
 
-        # Powershell adds in newline characters as it tries to wrap output around the display (by default 80 chars).
-        # This behavior is expected and cannot be changed, however it corrupts the XML e.g. newlines in the middle of
-        # element names; So instead, part of the XML is Base64 encoded prior to being put on STDOUT and in ruby all
-        # newline characters are stripped. Then where required decoded from Base64 back into text
-        out = REXML::Document.new(out.gsub(/\n/,""))
+        out[:exitcode] = out[:exitcode].to_i if !out[:exitcode].nil?
+        # if err contains data it must be "real" stderr output
+        # which should be appended to what PS has already captured
+        out[:stderr] = out[:stderr].nil? ? [] : [out[:stderr]]
+        out[:stderr] += err if !err.nil?
+        out[:native_stdout] = native_stdout
 
-        # picks up exitcode, errormessage, stdout and stderr
-        props = REXML::XPath.each(out, '//Property').map do |prop|
-          name = prop.attributes['Name']
-          value = (name == 'exitcode') ?
-            prop.text.to_i :
-            (prop.text.nil? ? nil : Base64.decode64(prop.text).force_encoding(Encoding::UTF_8))
-          # if err contains data it must be "real" stderr output
-          # which should be appended to what PS has already captured
-          if name == 'stderr'
-            value = value.nil? ? [] : [value]
-            value += err if !err.nil?
-          end
-          [name.to_sym, value]
-        end
-        props << [:native_stdout, native_stdout]
-
-        Hash[ props ]
+        out
       end
 
       def exit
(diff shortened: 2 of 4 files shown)


Iristyle:
The intermittent failure at https://ci.appveyor.com/project/puppetlabs/puppetlabs-powershell/build/1.1.x.376/job/t71njd7bnj4oxjcu is ... interesting.

Iristyle:
#144 can be merged prior to this ... or it can just be ignored.

lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
```diff
@@ -200,58 +229,115 @@ def self.wait_on(wait_object, timeout_ms = 50)
...
-        read_stdout(output_ready_event)
+        write_pipe(pipe_command(:execute))
+        write_pipe(pipe_data(powershell_code))
+        read_pipe(output_ready_event)
+      rescue EOFError => eoferror # a specific type of IOError
+        Puppet.warning [eoferror.inspect, eoferror.backtrace].join("\n")
+        # require 'pry'; binding.pry
+        # pipes are closed (could be as a result of tearing down process)
+        @usable = false
+        return nil, [eoferror.inspect]
```
glennsarti: This seems to be causing an issue later on.  This returns a single element array of type string however the code in https://github.com/puppetlabs/puppetlabs-powershell/blob/master/lib/puppet/provider/exec/powershell.rb#L83-L87 is expecting an array of an array of strings.  This is causing an error e.g.
```
#<NoMethodError: undefined method `each' for "#<EOFError: end of file reached>":String>
```

May have to change this ` return nil, [[eoferror.inspect]]` and the ones below as well

glennsarti: As per conversation. Perhaps change the warning to "Powershell Manager terminated unexpectedly" and add the backtrace to Debug.

Iristyle:
Builds on #144 which should be merged prior

glennsarti:
Ran this though a bunch of unless and onlyif tests.  Confirmed that if the PS process dies or a pipe failure, the resource behaves just like a non-zero exit code was returned from powershell.  No change in onlyif or unless behavior as part of this change,

Iristyle:
Some notes on this PR of a few outstanding bits based on review:

- [ ] tests to produce "real" stderr / stdout (native apps? cmd.exe?)
- [ ] probably shouldn't use gets in thread draining code since it relies on newline characters being sent (and will block until one is received) - if we instead use readpartial, we can glue together binary strings (even with split utf8 chars) and `.force_encoding` on the entire string to ensure its proper UTF8
- [ ] write tests to show that output ordering is interleaved correctly b/w Write-XXX / Console::Out (based on the `redirect stdout` commit)
- [ ] get rid of initializer params in PowerShellManager - @jpogran noted this makes it difficult to use - I concur
- [ ] rework the initialization to loop until pipe is ready - instead of waiting for a signal
    this is to improve cross-platform compat

ferventcoder:
Is this ready to be picked back up @Iristyle ?

spec/integration/puppet_x/puppetlabs/powershell_manager_spec.rb
```diff
@@ -173,6 +179,9 @@ def close_stream(stream, style = :inprocess)
       end
 
       it "should create a new PowerShell manager host if the output stream handle is closed" do
+        # currently skipped as it can trigger an internal Ruby thread clean-up race
+        # its unknown why this test fails, but not the identical test against @stderr
+        skip('This test can cause intermittent segfaults in Ruby with w32_reset_event invalid handle')
```
Iristyle: https://github.com/ruby/ruby/blob/ruby_2_3/lib/open3.rb#L201 is often where the error propagates from..

lib/puppet/provider/exec/powershell.rb
```diff
@@ -86,7 +90,7 @@ def run(command, check = false)
 
       Puppet.debug "STDERR: #{result[:errormessage]}" unless result[:errormessage].nil?
 
-      output = Puppet::Util::Execution::ProcessOutput.new(stdout.to_s || '', exit_code)
+      output = Puppet::Util::Execution::ProcessOutput.new(stdout.to_s + native_out.to_s, exit_code)
```
Iristyle: * `native_out` could be `nil`
* `stdout.to_s` was already in the code ... I think it too can be `nil` IIRC, but feel free to verify.

lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
```diff
@@ -33,18 +31,34 @@ def self.supported?
...
+        stdin.close
 
         Puppet.debug "#{Time.now} #{cmd} is running as pid: #{@ps_process[:pid]}"
 
-        init_ready_event_name =  "Global\\#{SecureRandom.uuid}"
-        init_ready_event = self.class.create_event(init_ready_event_name)
+        pipe_path = "\\\\.\\pipe\\#{named_pipe_name}"
+        # wait for the pipe server to signal ready, and fail if no response in 10 seconds
+
+        # wait up to 10 seconds in 0.2 second intervals to be able to open the pipe
```
Iristyle: Is 10 seconds long enough for PowerShell to startup and begin listening?  Or should we wait longer?

Iristyle:
I just repushed because I reran on PS2 / 2008 and realized that I used `StringBuilder.Clear()` again accidentally in what has been edited as 25e7032 to use `.Length = 0` instead.

jpogran:
This passes both CI and manual testing on win2008, win2012, and win2016 using PowerShell 2,3,4,5 and 5.1. LGTM

PR #143, opened 2016-09-09
(maint) Minor PS tweaks in prep for pipe rewrite

Iristyle:
- Pulls extraneous output from Invoke-PowerShellUserCode
- Refactors event signaling

### DIFF
--- a/lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
+++ b/lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
@@ -101,8 +101,9 @@ module PuppetX
       end
 
       def make_ps_init_code(init_ready_event_name)
+        debug_output = Puppet::Util::Log.level == :debug ? '-EmitDebugOutput' : ''
         <<-CODE
-. #{self.class.init_path} -InitReadyEventName '#{init_ready_event_name}'
+. #{self.class.init_path} -InitReadyEventName '#{init_ready_event_name}' #{debug_output}
         CODE
       end
 
--- a/lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
+++ b/lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
@@ -94,14 +94,16 @@ module PuppetX
         end
       end
 
-      def template_path
-        File.expand_path('../../../templates', __FILE__)
+      def self.init_path
+        path = File.expand_path('../../../templates', __FILE__)
+        path = File.join(path, 'init_ps.ps1').gsub('/', '\\')
+        "\"#{path}\""
       end
 
       def make_ps_init_code(init_ready_event_name)
-        template_file = File.new(template_path + "/init_ps.ps1.erb").read
-        template = ERB.new(template_file, nil, '-')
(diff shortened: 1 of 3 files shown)


lib/puppet_x/templates/init_ps.ps1.erb
```diff
@@ -501,19 +498,30 @@ function Invoke-PowerShellUserCode
...
+{
+  [CmdletBinding()]
+  param(
+    [String]
+    $EventName
+  )
+
+  $event = [System.Threading.EventWaitHandle]::OpenExisting($EventName)
+
+  [System.Diagnostics.Debug]::WriteLine("Signaling event $EventName")
```
Iristyle: Thoughts on whether to keep / remove the debug output?  This allows you to use SysInternals DbgView to monitor progress...

glennsarti: Sounds like the same problem with "Waited 50ms" debug log spam.  Perhaps jsut comment it out for the moment?

jpogran: Less so as you need to actively ask/listen for for those messages, its not output by default

Iristyle: It won't show up in any logs / for Puppet users.

If you just turn on DbgView, you'll see output from other apps flowing by (though that's usually turned on with a switch or config option)

glennsarti: Sorry...meant, same problem as; "How much debug output should we emit"?

glennsarti: I'm not really fussed either way.  If we keep it in, I would expect that we would properly instrument this with debug statements later.

jpogran:
👍

Iristyle:
FYI - I just added one more refactor commit (which is the basis for the next set of changes).  This changes how we initially bootstrap the manager, and removes ERB template stuff in favor of a static ps1 file.

lib/puppet_x/templates/init_ps.ps1.erb
Iristyle: I hid it behind `$DebugPreference` now.

lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
```diff
@@ -101,8 +101,9 @@ def self.init_path
       end
 
       def make_ps_init_code(init_ready_event_name)
+        debug_output = Puppet::Util::Log.level == :debug ? '$True' : '$False'
         <<-CODE
-. #{self.class.init_path} -InitReadyEventName '#{init_ready_event_name}'
+. #{self.class.init_path} -InitReadyEventName '#{init_ready_event_name}' -EmitDebugOutput #{debug_output}
```
Iristyle: @glennsarti I've given the user control over the output ... either via using `--debug` with Puppet or through `$DebugPreference`

Iristyle:
@puppetlabs/windows can I get a merge?

ferventcoder:
:+1:

PR #144, opened 2016-09-12
(MODULES-3875) Improve PowerShellManager resilience to failure

Iristyle:
- Existing test verified that 64k generated within PowerShell could
   be sent back to Ruby without deadlocking the Ruby parent process.

   Add a new test that originates 64k+ on the Ruby side and ensures
   that the data is properly round-tripped back to Ruby after
   passing it over stdin to PowerShell (which is implemented under
   the hood with popen3 via a named pipe).

- Previously it was possible to call exit against PowerShellManager
   and not have it completely terminate due to attempting to write to
   broken pipes, etc.

   Now verify state prior to attempting to close since Ruby will raise
   an IOError if calling close against an already closed IO instance

   Also ignore any errors raised when trying to 'exit' the instance

- Instead of having write_sdin / drain_pipe contain their own
   exception handlers, unify them under exec_read_result.

   This also correct two errors in how existing errors were raised:

   * The backtrace of the rescued exception was lost
   * The message would often include "Operation successfully completed"
     as the Windows::Error class is intended to be used after failures
     in FFI calls to Win32 APIs, which automatically call GetLastError
     to look up the reason for the API call failure.  However, since
     these errors are not the result of Windows API failures, the return
     value of GetLastErorr is 0 and the message is misleading.

   This will later allow other I/O and Pipe errors to be captured and
   handled in a single location and prevent code duplication.

- This new test documents how the PowerShell provider makes use of
   the PowerShellManager.  Specifically, it uses the .instance
   factory method and provides a command line.

   There were no existing tests that proved that the manager returns
   the same manager instance, bound to the same underlying PowerShell
   process - which more accureately represents how the provider uses
   the PowerShellManager

   These tests will be further expanded with new manager robustness
   code.

- Previously, an error that took out the PowerShell process could
   render the module unable to process any more PS commands - and
   anything else in the Puppet run using the module would fail.

   A simple way to reproduce this is by explicitly calling:

```
   [Diagnostics.Process]::GetCurrentProcess().Kill()
```

   This could also happen if a pipe breaks with EPIPE, or if any of
   the IO streams are closed to a bad file descriptor (EBADF) or a
   more generic IOError.

- Error handling for IO reads / writes is unified under the
   exec_read_result method. All EPIPE errors, IOError 'closed stream'
   and EBADF 'Bad file descriptor' errors are considered non-terminal,
   and simply inform the PowerShellManager that the current instance is
   unusable and the next time PowerShellManager.instance is called that
   a new PowerShell process should be created. It also ensures that the
   exit code from the execution is a -1, with the stderr containing the
   exception text.

- PowerShellManager.instance does this by explicitly calling an alive?
   helper which makes sure the process is running and all the streams
   are open and stat'able (stat was the only Ruby method that would
   generate an error when the underlying file descriptor or handle was
   bad).  When they are not, a new PowerShell process is created and
   this effectively self-heals between command execution.

- alive? is also used when reading the PowerShell stdout / stderr and
   waiting for PowerShell to have signaled that the output is ready.
   If this check is not performed, the PS process can die (as is the
   case with the new tests), prior to signaling the event that Ruby
   is waiting on, which deadlocks Puppet.

- Add a number of nefarious tests that tear down the underlying
   PowerShell process unexpectedly or that close the stdin, stdout or
   stderr streams in unexpected ways. Depending on Ruby version and
   the sequence of events these situations have slightly different
   failure modes which are captured in the tests.

### DIFF
--- a/lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
+++ b/lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
@@ -203,9 +203,6 @@ Invoke-PowerShellUserCode @params
 
       def write_stdin(input)
         @stdin.puts(input)
-      rescue => e
-        msg = "Error writing STDIN / reading STDOUT: #{e}"
-        raise Puppet::Util::Windows::Error.new(msg)
       end
 
       def drain_pipe(pipe, iterations = 10)
@@ -245,14 +242,14 @@ Invoke-PowerShellUserCode @params
         errors = errors.reject { |e| e.empty? }
 
         return output.join(''), errors
-      rescue => e
-        msg = "Error reading PIPE: #{e}"
-        raise Puppet::Util::Windows::Error.new(msg)
       end
 
       def exec_read_result(powershell_code, output_ready_event)
         write_stdin(powershell_code)
         read_stdout(output_ready_event)
+      rescue => e
+        msg = "Unexpected error with pipe communication"
+        raise e, msg, e.backtrace
       end
 
       if Puppet::Util::Platform.windows?
--- a/lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
+++ b/lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
@@ -72,10 +72,11 @@ module PuppetX
 
       def exit
         Puppet.debug "PowerShellManager exiting..."
-        @stdin.puts "\nexit\n"
-        @stdin.close
-        @stdout.close
-        @stderr.close
+        # ignore any failure to call exit against PS process
+        @stdin.puts "\nexit\n" if !@stdin.closed? rescue nil
+        @stdin.close if !@stdin.closed?
+        @stdout.close if !@stdout.closed?
+        @stderr.close if !@stderr.closed?
 
         exit_msg = "PowerShell process did not terminate in reasonable time"
         begin
--- a/lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
+++ b/lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
@@ -12,7 +12,15 @@ module PuppetX
       @@instances = {}
 
       def self.instance(cmd)
-        @@instances[:cmd] ||= PowerShellManager.new(cmd)
+        manager = @@instances[cmd]
+
+        if manager.nil? || !manager.alive?
+          # ignore any errors trying to tear down this unusable instance
+          manager.exit if manager rescue nil
+          @@instances[cmd] = PowerShellManager.new(cmd)
+        end
+
+         @@instances[cmd]
       end
 
       def self.win32console_enabled?
@@ -26,6 +34,7 @@ module PuppetX
       end
 
       def initialize(cmd)
+        @usable = true
         @stdin, @stdout, @stderr, @ps_process = Open3.popen3(cmd)
 
         Puppet.debug "#{Time.now} #{cmd} is running as pid: #{@ps_process[:pid]}"
@@ -41,6 +50,17 @@ module PuppetX
         at_exit { exit }
       end
 
+      def alive?
+        # powershell process running
+        @ps_process.alive? &&
+          # explicitly set during a read / write failure, like broken pipe EPIPE
+          @usable &&
+          # an explicit failure state might not have been hit, but IO may be closed
+          self.class.is_stream_valid?(@stdin) &&
+          self.class.is_stream_valid?(@stdout) &&
+          self.class.is_stream_valid?(@stderr)
+      end
+
       def execute(powershell_code, timeout_ms = nil, working_dir = nil)
         output_ready_event_name =  "Global\\#{SecureRandom.uuid}"
         output_ready_event = self.class.create_event(output_ready_event_name)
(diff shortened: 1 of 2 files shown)


jpogran:
tested works as advertisted 👍

Iristyle:
c820ff1 could arguably be a separate ticket - I encountered these fun issues while making sure the new pipe implementation behaved correctly.

I have some other pipe-breaking tests under dev that aren't in here yet, and they will build on some of the same ideas in that new test.

spec/integration/puppet_x/puppetlabs/powershell_manager_spec.rb
```diff
@@ -165,6 +169,24 @@ class PowerShellManager; end
       expect(first_pid).to eq(second_pid)
     end
 
+    it "should create a new PowerShell manager host if user code exits the first process" do
+      first_pid = manager.execute('[Diagnostics.Process]::GetCurrentProcess().Id')[:stdout]
+      exitcode = manager.execute('[Diagnostics.Process]::GetCurrentProcess().Kill()')[:exitcode]
+
+      # when a process gets torn down out from under manager, there is no response
+      # a nil exitcode is a reasonable signal that the PS process was torn out from under it
+      expect(exitcode).to eq(nil)
```
Iristyle: @jpogran / @ferventcoder / @glennsarti / @MosesMendoza - thoughts here on this?

If the underlying PS dies (whether intentionally or unintentionally crashes for instance) - then there will be no `exitcode` at all.  The provider should probably recognize this and behave accordingly at https://github.com/puppetlabs/puppetlabs-powershell/blob/master/lib/puppet/provider/exec/powershell.rb#L81

Any thoughts there?  What do other exec providers do if the process crashes?

Iristyle: Decided that we want the PowerShellManager to deal with unexpected pipe breakages and to return a `-1` error code with some `stderr` indicating bad things happened.

That allows us to proceed on subsequent resource applications.

Iristyle:
Should be merged prior to #142

lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
```diff
@@ -41,6 +50,19 @@ def initialize(cmd)
...
 
+      def alive?
+        # powershell process running
+        @ps_process.alive? &&
+          # explicitly set during a read / write failure, like broken pipe EPIPE
+          @usable &&
+          # an explicit failure state might not have been hit, but IO may be closed
+          # calling stat can yield an EBADF when underlying OS handle is bad
+          # TODO: what error is yielded when closed? is called on a bad pipe??
+          (!@stdin.closed? && !@stdin.stat.nil? rescue false) &&
```
Iristyle: I think instead of calling `.stat` here, I should see if a call to `get_osfhandle` is sufficient (my hunch is that it is)

lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
```diff
@@ -49,6 +71,9 @@ def execute(powershell_code, timeout_ms = nil, working_dir = nil)
 
         out, err = exec_read_result(code, output_ready_event)
 
+        # an error was caught during execution that has invalidated any results
+        return { :exitcode => -1, :stderr => err } if !@usable && out.nil?
```
Iristyle: This new failure type should be tested with `unless` and `onlyif` to make sure it works as we intend.

spec/integration/puppet_x/puppetlabs/powershell_manager_spec.rb
```diff
@@ -5,19 +5,230 @@
...
+
+        # TODO: necessary to find CRT version or just "msvcrt"
+        def self.msvcrt_name
+          host_part = RbConfig::CONFIG['host_os'].split("_")[1]
+          manifest  = File.join(RbConfig::CONFIG['bindir'], 'ruby.exe.manifest')
+
+          if host_part && host_part.to_i > 80 && File.exists?(manifest)
+            "msvcr#{host_part}"
+          else
+            "msvcrt"
```
Iristyle: I think just `msvcrt` is sufficient here - verify for certain to simplify this code.

glennsarti:
Note in the PR comments it says `A simple way to reproduce this is by explicitly calling:`  Should read;

```
A simple way to reproduce this is by explicitly calling:
exec { 'KillPS':
  command  => '([System.Diagnostics.Process]::GetCurrentProcess()).Kill()',
  provider  => powershell,
  logoutput => true,
 }
```

Iristyle:
@glennsarti that depends on context I think.  I was referring to the manager perspective, but I can add your manifest example for the user perspective as well.

lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
Iristyle: Proven incorrect... the CRT will still return a HANDLE value that was mapped to the file descriptor, without validating the HANDLE still exists or is otherwise valid.

Calling `stat` appears to be the best way to do this without additional FFI code

spec/integration/puppet_x/puppetlabs/powershell_manager_spec.rb
Iristyle: For posterity, the old method of finding `msvcrt` was the following:

```ruby
        def self.msvcrt_name
          host_part = RbConfig::CONFIG['host_os'].split("_")[1]
          manifest  = File.join(RbConfig::CONFIG['bindir'], 'ruby.exe.manifest')

          if host_part && host_part.to_i > 80 && File.exists?(manifest)
            "msvcr#{host_part}"
          else
            "msvcrt"
          end
        end
```

I don't believe this is necessary for the tests, and I'm updating.

PR #146, opened 2016-09-26
(maint) Merge master down to stable

jpogran:
Merge master down to stable in preparation for release


glennsarti:
There should be no further commits from master into stable for the release

PR #148, opened 2016-09-27
(MODULES-3443) Modify error test for differences in Powershell version

glennsarti:
Previously the integration tests for testing error behavior were passing on
Powershell 3+ however the error parsing is different in Powershell 2.0. This
commit modifies the test expectations for a line and char to returned but the
content of the line and char values are ignored.  This means the tests will now
pass on all Powershell versions but still ensure the correct behavior.

Currently the integration tests are failing on PowerShell version 2.0 or less.
This commit adds a helper function to determine the Powershell version without
the Powershell Manager code, and then skips the PowerShell Manager integration
tests if the PowerShell version is 2.0 or less.

Previously the working directory tests would fail if the path differed with casing however under
Windows paths are not case sensitive.  This commit modifies the test by changing
the string comparison to lowercase.

### DIFF
--- a/spec/integration/puppet_x/puppetlabs/powershell_manager_spec.rb
+++ b/spec/integration/puppet_x/puppetlabs/powershell_manager_spec.rb
@@ -472,7 +472,7 @@ $bytes_in_k = (1024 * 64) + 1
         result = manager.execute(powershell_runtime_error)
 
         expect(result[:exitcode]).to eq(1)
-        expect(result[:errormessage]).to match(/At line\:1 char\:33/)
+        expect(result[:errormessage]).to match(/At line\:\d+ char\:\d+/)
       end
     end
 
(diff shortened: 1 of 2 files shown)


spec/integration/puppet_x/puppetlabs/powershell_manager_spec.rb
```diff
@@ -31,8 +31,9 @@ module WindowsAPI
 end
 
 describe PuppetX::PowerShell::PowerShellManager,
-  :if => Puppet::Util::Platform.windows? && PuppetX::PowerShell::PowerShellManager.supported? do
-
+  :if => Puppet::Util::Platform.windows? && PuppetX::PowerShell::PowerShellManager.supported?,
```
Iristyle: This appears to have been because it wasn't well understood that the pipe race fix in #3144 should have been performed first.  After that merge, the only failing test (aside from the ones fixed in this PR) is the 64k test.  The only reason that test was failing was due to the difference in behavior between `Write-Output` in PS2 and PS3+ (PS2 introduces many extra newlines).

* #149 has been merged to address this concern.  That re-enabled 48 of the 49 tests.
* #150 reworked the tests so that all tests can be run regardless of PowerShell version

So at this point, we're good!

PR #149, opened 2016-09-28
(maint) Re-enable PowerShell 2 tests

Iristyle:
- In 5072e28 changes were made to exclude PowerShell 2 from running
   integration specs against PowerShellManager.  There are currently
   only 2 failures related to exceeding the Console BufferSize and
   how PowerShell 2 sends the output through PuppetPSHostUserInterface
   that actually need to be marked pending.

   Re-enable the 47 skipped tests under PowerShell 2 and mark the
   other 2 failing tests as pending.


glennsarti:
Will merge after running through adhoc.

PR #150, opened 2016-09-29
(maint) Fix multiline stderr capture

Iristyle:
- Previous code made two bad assumption that could result in the
   module crashing when anything is written to the stderr stream via
   native executables.

   The first assumption is that there is a non-nil value passed back in
   the XML response `stderr` element.  It's completely valid to have no
   pre-existing value present.

   The second assumption was that the captured stderr was a bare string.
   It will never be a bare string and should always be a single element
   string array at the very least.

 - Adjust the code to handle these situations accordingly and introduce
   a new test that creates multiline stderr output.

### DIFF
--- a/lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
+++ b/lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
@@ -89,7 +89,10 @@ module PuppetX
             (prop.text.nil? ? nil : Base64.decode64(prop.text))
           # if err contains data it must be "real" stderr output
           # which should be appended to what PS has already captured
-          value += err if err && (err != []) && (name == 'stderr')
+          if name == 'stderr'
+            value = value.nil? ? [] : [value]
+            value += err if !err.nil?
+          end
           [name.to_sym, value]
         end
 
--- a/lib/puppet/provider/exec/powershell.rb
+++ b/lib/puppet/provider/exec/powershell.rb
@@ -81,9 +81,7 @@ Puppet::Type.type(:exec).provide :powershell, :parent => Puppet::Provider::Exec 
       exit_code   = result[:exitcode]
 
       unless stderr.nil?
-        stderr.each do |er|
-          er.each { |e| Puppet.debug "STDERR: #{e.chop}" } unless er.empty?
-        end
+        stderr.each { |e| Puppet.debug "STDERR: #{e.chop}" unless e.empty? }
       end
 
       Puppet.debug "STDERR: #{result[:errormessage]}" unless result[:errormessage].nil?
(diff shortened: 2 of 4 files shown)


lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
```diff
@@ -89,7 +89,7 @@ def execute(powershell_code, timeout_ms = nil, working_dir = nil)
             (prop.text.nil? ? nil : Base64.decode64(prop.text))
           # if err contains data it must be "real" stderr output
           # which should be appended to what PS has already captured
-          value += err if err && (err != []) && (name == 'stderr')
+          value = (value || '') + err.join('') if err && (err != []) && (name == 'stderr')
```
ferventcoder: No more `+=`

ferventcoder: Completely understand why though - to ensure that the value is non-nil. 

It looks like there is a possibility of `value.nil?` being true still. Were you trying to remove this scenario so that it is at least always an empty string?

ferventcoder: If `prop.text.nil?`, value remains nil from lines 87-89.
If `err` is nil or the name doesn't equal stderr.

ferventcoder: Is the check for `err != []` still necessary?

Iristyle: I made a few more changes there after some more testing - would appreciate some more testing!

lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
```diff
@@ -89,7 +89,10 @@ def execute(powershell_code, timeout_ms = nil, working_dir = nil)
             (prop.text.nil? ? nil : Base64.decode64(prop.text))
           # if err contains data it must be "real" stderr output
           # which should be appended to what PS has already captured
-          value += err if err && (err != []) && (name == 'stderr')
+          if name == 'stderr'
+            value = value.nil? ? [] : [value]
+            value += err if !err.nil?
```
ferventcoder: no more join then?

Iristyle: I made notes in commit message... our previous behavior was to pass an array of strings up through "stderr" - but I accidentally broke it in fa3d618.  There were no tests in the provider to catch it.

We should probably rethink that at some point (to just use a string value everywhere, like we do for `stdout`) - but I'm trying to preserve what we were doing.

ferventcoder: Makes sense.

PR #153, opened 2016-10-10
DOC-2960: new limitation and a bit of editing

jbondpdx:
Added a thing about here-strings. Hit up a few editorial and formatting fixes while I was in there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -197,7 +197,16 @@ Runs the `exec`, unless the command returns 0. Valid options: String. Default: U
 
 * Only supported on Powershell 2.0 and above.
 
-* To use here-strings in scripts for this module, you must use the `@" / "@` syntax in your code.
+* When using here-strings in inline or templated scripts executed by this module, you must use the double-quote style syntax that begins with `@"` and ends with `"@`. The single-quote syntax that begins with `@'` and ends with `'@` is not supported.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -120,65 +123,82 @@ exec { 'test':
...
 
 ## Limitations
 
- * Only supported on Windows Server 2008 and above, and Windows 7 and above.
+* Only supported on Windows Server 2008 and above, and Windows 7 and above.
+
+* Only supported on Powershell 2.0 and above.
 
- * Only supported on Powershell 2.0 and above.
+* To use here-strings in scripts for this module, you must use the `@" / "@` syntax in your code.
```
Iristyle: When using here-strings within inline or templated scripts executed by this module, you must use the double-quote style syntax that begins with `@"` and ends with `"@`.  The single-quote syntax that begins with `@'` and ends with `'@` is not supported.

Note that any external `.ps1` script file loaded or executed with the call operator `&` is not subject to this limitation, and may contain any style here-string. For instance, the script file `external-code.ps1` may contain any style here-string:

~~~ puppet
exec { 'external-code':
  command   => '& C:\external-code.ps1',
  provider  => powershell,
}
~~~

jbondpdx: Awesome, thanks @Iristyle ! Changes are made and pushed.

jpogran:
LGTM

PR #155, opened 2016-10-31
(MODULES-3690) PowerShell v2 + .NET Framework less than 3.5 is unsupported for PowerShell Manager

ferventcoder:
Determine if the PowerShell version installed is compatible with module. When folks are using PowerShell 3+, they are on .NET Framework 4+, so they are completely compatible with the PowerShell Manager. When folks are using PowerShell v2 and .NET 3.5, they are also compatible, as `System.IO.Pipes` requires at least .NET Framework 3.5.

However if they are running PowerShell v2 with .NET 2.0 (which is not going to happen very often, but it represents a base install on Windows Server 2008 and R2), it is not compatible with the PowerShell Manager. Ensure it switches back to v1 functionality.

Note:
This brings the ability to detect the PowerShell version over from the PowerShell DSC module. I adjusted it to use `PuppetX::PuppetLabs::PowerShell::PowerShellVersion` instead of just `PuppetX::PuppetLabs::PowerShellVersion` like the DSC module to be more proper and avoid conflicts.

### DIFF
--- a/lib/puppet_x/puppetlabs/powershell/compatible_powershell_version.rb
+++ b/lib/puppet_x/puppetlabs/powershell/compatible_powershell_version.rb
@@ -1,0 +1,51 @@
+require File.join(File.dirname(__FILE__), 'powershell_version')
+
+module PuppetX
+  module PuppetLabs
+    module PowerShell
+      class CompatiblePowerShellVersion
+        def self.compatible_version?
+          value = false
+
+          powershell_version = PuppetX::PuppetLabs::PowerShell::PowerShellVersion.version
+
+          return false if powershell_version.nil?
+
+          # PowerShell v1 - definitely not good to go. Really the entire module
+          # may not even work but I digress
+          return false if Gem::Version.new(powershell_version) < Gem::Version.new(2)
+
+          # PowerShell v3+, we are good to go b/c .NET 4+
+          # https://msdn.microsoft.com/en-us/powershell/scripting/setup/windows-powershell-system-requirements
+          # Look at Microsoft .NET Framwork Requirements section.
+          if Gem::Version.new(powershell_version) >= Gem::Version.new(3)
+            return true
+          end
+
+          # If we are using PowerShell v2, we need to see what the latest
+          # version of .NET is that we have
+          # https://msdn.microsoft.com/en-us/library/hh925568.aspx
+          if Puppet::Util::Platform.windows?
+            require 'win32/registry'
+
+            begin
+              # At this point in the check, PowerShell is using .NET Framework
+              # 2.x family, so we only need to verify v3.5 key exists.
+              # If we were verifying all compatible types we would look for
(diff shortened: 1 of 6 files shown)


lib/puppet_x/puppetlabs/powershell/powershell_version.rb
```diff
@@ -0,0 +1,53 @@
...
+            rescue
+              version = nil
+            end
+            version
+          end
+        end
+      end
+    end
+  end
+end
```
ferventcoder: This comes from DSC.

spec/unit/puppet_x/puppetlabs/powershell/powershell_version_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+    it "should return nil and not throw" do
+      Win32::Registry.any_instance.expects(:open).with('SOFTWARE\Microsoft\PowerShell\3\PowerShellEngine', Win32::Registry::KEY_READ | 0x100).raises(Win32::Registry::Error.new(2), 'nope').once
+      Win32::Registry.any_instance.expects(:open).with('SOFTWARE\Microsoft\PowerShell\1\PowerShellEngine', Win32::Registry::KEY_READ | 0x100).raises(Win32::Registry::Error.new(2), 'nope').once
+
+      version = @ps.version
+
+      expect(version).to eq nil
+    end
+  end
+end
```
ferventcoder: This comes from the DSC module.

lib/puppet_x/puppetlabs/powershell/compatible_powershell_version.rb
```diff
@@ -0,0 +1,50 @@
...
+              # If we were verifying all compatible types we would look for
+              # any of the following:
+              # - v3.5 key
+              # - then v4.0
+              # - then v4
+              hive = Win32::Registry::HKEY_LOCAL_MACHINE
+              # redirection doesn't actually matter here - disable it anyway
+              hive.open('SOFTWARE\Microsoft\NET Framework Setup\NDP\v3.5', Win32::Registry::KEY_READ | 0x100) do |reg|
+                value = true
+              end
```
ferventcoder: There is a teeny, tiny possibility that someone has went in an configured PowerShell.exe.config to use .NET framework v4 when on PowerShell v2. Not many people are going to do this, as it is pretty much an unsupported scenario. 

This could cause an issue if they have .NET 4.x installed but also only .NET 2.0. There is almost zero chance this will ever be hit, so it's really not worth the check at all.

glennsarti: ~~So we don't support .Net 3.0 either?~~  .Net 3.0 is no longer supported.  3.5 SP1 is the supported version in the 3.x family.

ferventcoder:
@puppetlabs/windows @puppetlabs/modules this is ready for review.

PR #156, opened 2016-11-07
Fixes Powershell (1.0) not executing on Windows Server 2016 Standard

rdev5:
Powershell module does not work on Windows Server 2016 Standard (Windows 10.0). This pull request makes the necessary "changes" to restore functionality and is backwards compatible with Windows Server 2012 R2 Standard.

### DIFF
--- a/lib/puppet/provider/exec/powershell.rb
+++ b/lib/puppet/provider/exec/powershell.rb
@@ -65,7 +65,7 @@ Puppet::Type.type(:exec).provide :powershell, :parent => Puppet::Provider::Exec 
         # we redirect powershell's stdin to read from the file. Current
         # versions of Windows use per-user temp directories with strong
         # permissions, but I'd rather not make (poor) assumptions.
-        return super("cmd.exe /c \"\"#{native_path(command(:powershell))}\" #{legacy_args} -Command - < \"#{native_path}\"\"", check)
+        return super("\"#{native_path(command(:powershell))}\" #{legacy_args} -Command \"Get-Content \\\"#{native_path}\\\" | Invoke-Expression\"")
       end
     else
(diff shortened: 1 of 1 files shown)


ferventcoder:
@rdev5 did you mean to target the 1.x branch here?

ferventcoder:
or rather, was it an accident that you created the PR for the master branch?

rdev5:
@ferventcoder While 1.x is the affected version, the proposed changes would also apply to the "fallback" implementation where [`PowershellManager` was not supported](https://github.com/puppetlabs/puppetlabs-powershell/blob/master/lib/puppet/provider/exec/powershell.rb#L57) on [Line 68](https://github.com/puppetlabs/puppetlabs-powershell/blob/master/lib/puppet/provider/exec/powershell.rb#L68)

ferventcoder:
@rdev5 while it's true that the currently released powershell module doesn't work on Windows Server 2016 due to a bug in PowerShell 5.1, we've already fixed stable and will be releasing a version that does work in Windows Server 2016. Can you give that a shot and let us know if it meets your needs?

https://tickets.puppetlabs.com/browse/MODULES-3690 for reference.

ferventcoder:
@rdev5 are these changes still necessary now that we've merged and prepared to release 2.1.0?

rdev5:
@ferventcoder I'll try to do some testing sometime tomorrow. Our patch sufficed our needs in restoring functionality so we just sort of ran with that up until now. Will let you know. Thanks!

jpogran:
@rdev5 master branch has supported WMF 5.1 for awhile now, as well as the December update of WMF5.1 fixes for the old way the code worked. Thank for the contribution, but this PR is no longer needed.

PR #159, opened 2016-11-15
(MODULES-3632) Update modulesync

hunner:
https://github.com/puppetlabs/modulesync_configs/pull/99


hunner:
and also https://github.com/puppetlabs/modulesync_configs/pull/103

PR #168, opened 2017-01-06
(MODULES-4263) add blacksmith rake tasks

eputnam:
new CI changes require that windows modules have access to blacksmith tasks


ferventcoder:
Should you start from stable on some of these?

PR #177, opened 2017-04-20
(DO NOT MERGE) Ask ps_manager to exit after catalog

Iristyle:
@puppetlabs/windows quick spike on the `post_resource_eval` implementation

### DIFF
--- a/lib/puppet/provider/exec/powershell.rb
+++ b/lib/puppet/provider/exec/powershell.rb
@@ -115,6 +115,13 @@ Puppet::Type.type(:exec).provide :powershell, :parent => Puppet::Provider::Exec 
     true
   end
 
+  def self.post_resource_eval
+    # TODO: make sure this isn't done in --noop and is only done when manager has been created
+    if PuppetX::PowerShell::PowerShellManager.supported?
+      ps_manager.exit
+    end
+  end
+
   private
   def write_script(content, &block)
     Tempfile.open(['puppet-powershell', '.ps1']) do |file|
--- a/lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
+++ b/lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
@@ -108,6 +108,10 @@ module PuppetX
         @usable = false
 
         Puppet.debug "PowerShellManager exiting..."
+
+        # ask server to shutdown if its still running
+        write_pipe(pipe_command(:exit)) if !@pipe.closed?
+
         # pipe may still be open, but if stdout / stderr are dead PS process is in trouble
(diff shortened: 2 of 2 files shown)


lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
```diff
@@ -108,6 +108,10 @@ def exit
         @usable = false
 
         Puppet.debug "PowerShellManager exiting..."
+
```
Iristyle: Could plausibly make sense to just remove the `at_exit` handling altogether (which is caused some crashes at the end of the AppVeyor run because of `exit` being called for a second time).

lib/puppet/provider/exec/powershell.rb
```diff
@@ -115,6 +115,13 @@ def validatecmd(command)
     true
   end
 
+  def self.post_resource_eval
+    # TODO: make sure this isn't done in --noop and is only done when manager has been created
+    if PuppetX::PowerShell::PowerShellManager.supported?
```
glennsarti: Hrmm how do we know if manager "exists"?

lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
```diff
@@ -116,7 +120,7 @@ def exit
         @stderr.close if !@stderr.closed?
 
         # wait up to 2 seconds for the watcher thread to fully exit
-        @ps_process.join(2)
+        @ps_process.join(2) if @ps_process.alive?
```
glennsarti: Should we be iterating over `@@instances` and killing all instances not just this one?

Should there be a different method to `exit_all` ?

lib/puppet/provider/exec/powershell.rb
Iristyle: Could just be the addition of an instance variable for the provider instance indicating a manager has been created (i.e. that `run` happened)

lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
Iristyle: I was going to say no, because:

`exit` is an instance method.  Each instance, upon creation, registers an `at_exit` handler to call `exit` for its instance.

The provider instances each ask for a `ps_manager` and then in `post_resource_eval` should clean it up... 

BUT

I just noticed that its `self.post_resource_eval`, so there should never be a `ps_manager` instance in scope.  In other words, `post_resource_eval` runs once for a type of provider, not per instance of provider.

So I may have to rethink this a bit... thanks!

Iristyle:
Rolling with the simpler #179 for now

PR #178, opened 2017-04-21
(SPIKE)(MODULES-4754) Incomplete hard teardown of powershell.exe

Iristyle:
This is totally incomplete and possibly worth abandoning depending on if MODULES-4754 can be easily repro'd.

The idea here would be to more forcefully guarantee a PowerShell process should be terminated.... but it would use Windows APIs and therefore only work on Windows.

### DIFF
--- a/lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
+++ b/lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
@@ -108,6 +108,10 @@ module PuppetX
         @usable = false
 
         Puppet.debug "PowerShellManager exiting..."
+
+        # ask server to shutdown if its still running
+        write_pipe(pipe_command(:exit)) if !@pipe.closed?
+
         # pipe may still be open, but if stdout / stderr are dead PS process is in trouble
         # and will block forever on a write to the pipe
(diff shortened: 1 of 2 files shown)


lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
```diff
@@ -46,6 +188,7 @@ def initialize(cmd, debug)
         ps_args = ['-File', self.class.init_path, "\"#{named_pipe_name}\""]
         ps_args << '"-EmitDebugOutput"' if debug
         # @stderr should never be written to as PowerShell host redirects output
+        # TODO: take a fd and turn it into a handle
         stdin, @stdout, @stderr, @ps_process = Open3.popen3("#{cmd} #{ps_args.join(' ')}")
```
Iristyle: A better way to do this on Windows would be to start `powershell.exe` through `Process.create` which is monkey-patched by `win32-process` gem, allowing immediate access to the real handle.

However, that puts the module back into the mode of supporting multiple OSes differently, which is less than ideal.  (`popen3` uses Rubys `Process.spawn` under the hood)

Iristyle:
Related to #177

PR #179, opened 2017-04-21
(MODULES-4754) Dispose runspace on pipe close

Iristyle:
- In the finally block of the Start-PipeServer listener, explicitly
   Dispose of the current loaded PowerShell runspace.


Iristyle:
Related to #178 and #177

glennsarti:
@jpogran Any thoughts on this?

PR #193, opened 2017-07-25
(MODULES-5228) Move PowerShell Manager template file

glennsarti:
Previously the PowerShell manager template file lived in the root of the
templates directory, however so does the same file in the DSC module which has
caused issues as they diverge.  This commit moves the template file to a
PowerShell module specific location so that it should not conflict with any
other modules.


Iristyle:
FWIW, I think we should try and namespace all helper code under `puppet_x/puppetlabs/powershell`, like

```
puppet_x/puppetlabs/powershell/powershell_manager.rb
puppet_x/puppetlabs/powershell/init_ps1.ps1
```

Rather than introducing separate namespacing like this PR has done

```
puppet_x/puppetlabs/powershell/powershell_manager.rb
puppet_x/puppetlabs/templates/powershell/init_ps1.ps1
```

PR #200, opened 2017-11-21
(MODULES-6081) Rename PowerShell executable

RandomNoun7:
This commit renames the PowerShell binary name from PowerShell to
pwsh. This is because the name of the binary has changed in the
latest PowerShell betas.

### DIFF
--- a/lib/puppet/provider/exec/powershell.rb
+++ b/lib/puppet/provider/exec/powershell.rb
@@ -8,12 +8,8 @@ Puppet::Type.type(:exec).provide :powershell, :parent => Puppet::Provider::Exec 
       "#{ENV['SYSTEMROOT']}\\sysnative\\WindowsPowershell\\v1.0\\powershell.exe"
     elsif File.exists?("#{ENV['SYSTEMROOT']}\\system32\\WindowsPowershell\\v1.0\\powershell.exe")
       "#{ENV['SYSTEMROOT']}\\system32\\WindowsPowershell\\v1.0\\powershell.exe"
-    elsif File.exists?('/usr/bin/powershell')
-      '/usr/bin/powershell'
-    elsif File.exists?('/usr/local/bin/powershell')
-      '/usr/local/bin/powershell'
     elsif !Puppet::Util::Platform.windows?
-      "powershell"
+      'pwsh'
     else
       'powershell.exe'
     end
--- a/spec/unit/provider/exec/powershell_spec.rb
+++ b/spec/unit/provider/exec/powershell_spec.rb
@@ -31,12 +31,8 @@ describe Puppet::Type.type(:exec).provider(:powershell) do
       "#{ENV['SYSTEMROOT']}\\sysnative\\WindowsPowershell\\v1.0\\powershell.exe"
     elsif File.exists?("#{ENV['SYSTEMROOT']}\\system32\\WindowsPowershell\\v1.0\\powershell.exe")
       "#{ENV['SYSTEMROOT']}\\system32\\WindowsPowershell\\v1.0\\powershell.exe"
-    elsif File.exists?('/usr/bin/powershell')
-      '/usr/bin/powershell'
-    elsif File.exists?('/usr/local/bin/powershell')
-      '/usr/local/bin/powershell'
     elsif !Puppet::Util::Platform.windows?
-      "powershell"
+      'pwsh'
     else
       'powershell.exe'
     end


lib/puppet/provider/exec/powershell.rb
```diff
@@ -8,12 +8,10 @@
       "#{ENV['SYSTEMROOT']}\\sysnative\\WindowsPowershell\\v1.0\\powershell.exe"
     elsif File.exists?("#{ENV['SYSTEMROOT']}\\system32\\WindowsPowershell\\v1.0\\powershell.exe")
       "#{ENV['SYSTEMROOT']}\\system32\\WindowsPowershell\\v1.0\\powershell.exe"
-    elsif File.exists?('/usr/bin/powershell')
-      '/usr/bin/powershell'
-    elsif File.exists?('/usr/local/bin/powershell')
-      '/usr/local/bin/powershell'
-    elsif !Puppet::Util::Platform.windows?
```
glennsarti: The default for non-windows platforms is needed because linux etc. won't have the file extension.

spec/unit/provider/exec/powershell_spec.rb
```diff
@@ -31,12 +31,10 @@
...
-      '/usr/bin/powershell'
-    elsif File.exists?('/usr/local/bin/powershell')
-      '/usr/local/bin/powershell'
-    elsif !Puppet::Util::Platform.windows?
-      "powershell"
+    elsif File.exists?('/usr/bin/pwsh')
+      '/usr/bin/pwsh'
+    elsif File.exists?('/usr/local/bin/pwsh')
+      '/usr/local/bin/pwsh'
     else
```
glennsarti: Same as above.

lib/puppet/provider/exec/powershell.rb
Iristyle: You beat me to it...

RandomNoun7: True. I think there was some question about when this location should ever be reached, but I guess if someone has deliberately installed to weird directory and put in `PATH` it would get here.

lib/puppet/provider/exec/powershell.rb
```diff
@@ -8,12 +8,12 @@
       "#{ENV['SYSTEMROOT']}\\sysnative\\WindowsPowershell\\v1.0\\powershell.exe"
     elsif File.exists?("#{ENV['SYSTEMROOT']}\\system32\\WindowsPowershell\\v1.0\\powershell.exe")
       "#{ENV['SYSTEMROOT']}\\system32\\WindowsPowershell\\v1.0\\powershell.exe"
-    elsif File.exists?('/usr/bin/powershell')
-      '/usr/bin/powershell'
-    elsif File.exists?('/usr/local/bin/powershell')
-      '/usr/local/bin/powershell'
+    elsif File.exists?('/usr/bin/pwsh')
```
Iristyle: I haven't looked at this in a while... shouldn't the symlinks in `/usr/bin` and `/usr/local/bin` to `pwsh` already be in `PATH` on non-Windows machines? Is there a downside to observing a preference for `/usr/bin` over `/usr/local/bin`?

On non-Windows platforms, typically `/usr/local/bin` appears in `PATH` *before* `/usr/bin`, giving it preferential treatment.

glennsarti: If we can do command detection via function that would be preferred i.e. use Puppet's where helper to find pwsh.  Didn't think we could.

RandomNoun7: I was just wondering if there was a Ruby native way to search each dir in `PATH` to find the right one.

Iristyle: My point was that we shouldn't have to "find `pwsh`" at all, should we?

Given `/usr/local/bin` and `/usr/bin` are already in `PATH` (and in the right preference order already), shouldn't this be sufficient?

```
elsif !Puppet::Util::Platform.windows?
  'pwsh'
```

RandomNoun7: So given that we talked about the possibility of a process account not having `/usr/local/bin` in `PATH`, I propose that I change the order of the block so that `/usr/local/bin` is before `usr/bin`, and then if neither works, we can just call `pwsh` on the off chance that they have it placed somewhere else for some reason.

Iristyle: I think in this case we should prefer to not replicate the shell's resolution behavior, or guess the appropriate ordering, and just use the "assume it's in PATH approach" - which is the norm inside of Puppet provider suitability checks already.

Take the user provider for instance - all of the commands are checked for existence, but no paths are qualified - https://github.com/puppetlabs/puppet/blob/master/lib/puppet/provider/user/useradd.rb#L12

I spot checked Ubuntu and see that `root` (default agent service account) has `/usr/local/bin` in `PATH` already... and it looks like the PowerShell packages MS is providing drop a symlink in `/usr/bin` anyhow. We could check the other platforms where  PowerShell runs to be certain, but my hunch is there shouldn't be any problems resolving an unqualified `pwsh`

RandomNoun7: On my mac the symlink is in `/usr/local/bin`, but same difference.

RandomNoun7: The latest commit does the 32bit checking on Windows, but for non-Windows just lets the platform sort it out based on `PATH`.

Iristyle: Yeah, OSX has a preference for `/usr/local/bin` over `/usr/bin` - you can double check the service user's account if you want to make sure it's in `PATH` by default.

RandomNoun7: Either way though, that's another reason to defer to the platform behavior, so that each platform user gets their idiomatic expected results.

PR #201, opened 2017-11-21
(MODULES-4748) Increase timeout for opening PowerShell

michaeltlombardi:
Prior to this commit, the timeout waiting for PowerShell to start
was only 10 seconds; this was not, in practice, a long enough
period of time for multiple customers, especially when a node is
under heavy load. This commit increases the timeout to 30 seconds.


Iristyle:
Unrelated TravisCI failures fixed in #200

PR #202, opened 2017-11-22
(MODULES-4748) Prevent zombie procs from timeout

michaeltlombardi:
Prior to this commit if a timeout occurs while waiting for the
connection to a named pipe then a zombie process is left on the
system as the code to close or kill it is never called. This commit
ensures that such zombie processes are removed before the error is
raised.

### DIFF
--- a/lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
+++ b/lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
@@ -9,14 +9,14 @@ module PuppetX
     class PowerShellManager
       @@instances = {}
 
-      def self.instance(cmd, debug = false)
+      def self.instance(cmd, debug = false, pipe_timeout = 30)
         key = cmd + debug.to_s
         manager = @@instances[key]
 
         if manager.nil? || !manager.alive?
           # ignore any errors trying to tear down this unusable instance
           manager.exit if manager rescue nil
-          @@instances[key] = PowerShellManager.new(cmd, debug)
+          @@instances[key] = PowerShellManager.new(cmd, debug, pipe_timeout)
         end
 
          @@instances[key]
@@ -38,7 +38,7 @@ module PuppetX
         !win32console_enabled?
       end
 
-      def initialize(cmd, debug)
+      def initialize(cmd, debug, pipe_timeout)
         @usable = true
 
         named_pipe_name = "#{SecureRandom.uuid}PuppetPsHost"
@@ -55,18 +55,29 @@ module PuppetX
         # wait for the pipe server to signal ready, and fail if no response in 10 seconds
 
         # wait up to 30 seconds in 0.2 second intervals to be able to open the pipe
-        150.times do
+        # If the pipe_timeout is ever specified as less than the sleep interval it will
+        # never try to connect to a pipe and error out as if a timeout occurred.
+        sleep_interval = 0.2
+        (pipe_timeout / sleep_interval).to_int.times do
           begin
             # pipe is opened in binary mode and must always
             @pipe = File.open(pipe_path, 'r+b')
             break
           rescue
-            sleep 0.2
+            sleep sleep_interval
           end
         end
-
-        fail "Failure waiting for PowerShell process #{@ps_process[:pid]} to start pipe server" if @pipe.nil?
-
+        if @pipe.nil?
+          # Tear down and kill the process if unable to connect to the pipe; failure to do so
+          # results in zombie processes being left after the puppet run. We discovered that
+          # Closing @ps_process via .kill instead of using this method actually kills the watcher
+          # and leaves an orphaned process behind. Failing to close stdout and stderr also leaves
+          # clutter behind, so explicitly close those too.
+          @stdout.close if !@stdout.closed?
+          @stderr.close if !@stderr.closed?
+          Process.kill("KILL", @ps_process[:pid]) if @ps_process.alive?
+          raise "Failure waiting for PowerShell process #{@ps_process[:pid]} to start pipe server"
+        end
         Puppet.debug "#{Time.now} PowerShell initialization complete for pid: #{@ps_process[:pid]}"
 
         at_exit { exit }
--- a/spec/integration/puppet_x/puppetlabs/powershell_manager_spec.rb
+++ b/spec/integration/puppet_x/puppetlabs/powershell_manager_spec.rb
@@ -40,7 +40,7 @@ describe PuppetX::PowerShell::PowerShellManager,
     "#{powershell} #{cli_args.join(' ')}"
   }
 
-  def create_manager
+  def create_manager()
     PuppetX::PowerShell::PowerShellManager.instance(manager_args, true)
   end
 
@@ -56,6 +56,18 @@ describe PuppetX::PowerShell::PowerShellManager,
 
         expect(manager_2).to eq(manager)
         expect(first_pid).to eq(second_pid)
+      end
+
+      it "should fail if the manger is created with a short timeout" do
+        expect {
+          PuppetX::PowerShell::PowerShellManager.new(manager_args, false, 0.01)
+        }.to raise_error do |e|
+          expect(e).to be_a(RuntimeError)
+          expected_error = /Failure waiting for PowerShell process (\d+) to start pipe server/
+          expect(e.message).to match expected_error
+          pid = expected_error.match(e.message)[1].to_i
+          expect{Process.kill(0, pid)}.to raise_error(Errno::ESRCH)
+        end
       end
 
       def bad_file_descriptor_regex


michaeltlombardi:
We'll need to add a test to verify that the timeout code triggers. I verified manually by setting the timeout to `0.2` seconds. I'm not sure this is the most elegant way to close up the process either, I'm open to suggestions/improvements.

lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
```diff
@@ -55,7 +55,7 @@ def initialize(cmd, debug)
         # wait for the pipe server to signal ready, and fail if no response in 10 seconds
 
         # wait up to 30 seconds in 0.2 second intervals to be able to open the pipe
-        150.times do
+        (pipe_timeout * 5).times do
```
Iristyle: Not sure if this makes sense, but maybe...

```ruby 
sleep_interval = 0.2
(pipe_timeout / sleep_interval).times do

# ...

sleep sleep_interval
```

michaeltlombardi: Does `times` work with a float? I got errors when I tried but didn't investigate too closely in an earlier local iteration of this.

RandomNoun7:
Aren't we doing to need to change a few other method signatures to expose this far enough down the stack to be able to test it? I have something similar to this and I can get tests to pass and fail by restricting the wait time, but I don't have this all traced out far enough to be able to set that timeout from an rspec test call.

lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
RandomNoun7: Float doesn't have the .times method. Only int. You could do `(pipe_timeout / sleep_interval).to_i.times` and that will work with some imprecision to avoid errors.

lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
```diff
@@ -64,9 +64,14 @@ def initialize(cmd, debug)
...
         end
-
-        fail "Failure waiting for PowerShell process #{@ps_process[:pid]} to start pipe server" if @pipe.nil?
-
+        if @pipe.nil?
+          # Tear down and kill the process if unable to connect to the pipe; failure to do so
+          # results in zombie processes being left after the puppet run.
+          @stdout.close if !@stdout.closed?
+          @stderr.close if !@stderr.closed?
+          Process.kill("KILL", @ps_process[:pid])
```
Iristyle: So one question that arises is... should we grab the process handle immediately after starting the process, as the system may reuse PIDs (assuming `@ps_process` doesn't already have a handle)? If the process fails to start in the first place or quickly crashes and burns, we've now waited 30 seconds, which increases the likelihood that the PID has been reassigned.  That said, PIDs cannot be reused until all handles to the process are closed - and I'm not sure what state Ruby is actually holding on to at this point - so we may have to look to the Ruby source for clarity.

We should be able to use an API call to `OpenProcess` with `PROCESS_ALL_ACCESS`, `TRUE`, `@ps_process[:pid]` to get a handle, which we can then use to kill the process here... but that may be overkill (pardon the pun)

Iristyle: `popen3` returns the same return type as [Process.detach](https://ruby-doc.org/core-2.1.1/Process.html#method-c-detach) at https://github.com/ruby/ruby/blob/trunk/lib/open3.rb#L200

The nominal implementation of `Process.detach` lives in the Ruby source at https://github.com/ruby/ruby/blob/fe3decb2017af2092d4b2181e27e4cd8f6bbf4fa/process.c#L1086-L1093

The Ruby code gets a little squirrely though as the implementations here use `waitpid` and `wait4`, but I believe the Windows variant is backed by the code at https://github.com/ruby/ruby/blob/e33b1690d06f867e45750bd8e3e8b06d78b5bc26/win32/win32.c#L4496-L4569. You'll also find the expected calls to `GetExitCodeProcess` here.... and each of those calls seems to be relying on `child->hProcess`.  That makes me think Ruby is holding on to a handle, even if it isn't exposed in its APIs directly anywhere.

Iristyle: So after verifying while pairing, this seems correct.

One thing that we do need to change slightly is to make:

```ruby
Process.kill("KILL", @ps_process[:pid]) if @ps_process.alive?
```

This ensures that the waiter thread is still up and monitoring the child process, which is a proxy for the child process still running.  If the waiter thread is *dead*, there's no guarantee the PID is even valid, or that another process hasn't claimed it.

We want to be careful about introducing the "PowerShell module chaos monkey" on unsuspecting victims.

michaeltlombardi:
@Iristyle: Updated with the addition of a test to verify timeouts by @RandomNoun7. To do so we had to fake a _very_ low timeout (lower than the `0.2` that was sufficient in our previous explorations). I *just realized* we probably should've adjusted my commit to include the changes to `self.instance` - we discovered that, unless we add it there, we couldn't find a way to pass a pipe timeout to the initialize block.

This could overall probably use some cleanup, but it feels like a good break point for some introspection.

lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
```diff
@@ -55,7 +55,7 @@ def initialize(cmd, debug, pipe_timeout = 30)
         # wait for the pipe server to signal ready, and fail if no response in 10 seconds
 
         # wait up to 30 seconds in 0.2 second intervals to be able to open the pipe
-        sleep_interval = 0.2
+        sleep_interval = 0.01
         (pipe_timeout / sleep_interval).to_int.times do
```
RandomNoun7: This match ```(pipe_timeout / sleep_interval).to_int``` comes out to zero if ```pipe_timeout``` is ever allowed to get to this point and has a value less than ```sleep_interval```. Is this something we feel we can ignore, or do want to test and throw a descriptive error if that occurs? This risk here is that since external users won't usually plumb the code deep enough to know what a safe value is, they could potentially pass anything in here as a value.

Iristyle: Puppet code should be the only caller here, so I don't *think* we'll need to be overly defensive.

lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
```diff
@@ -55,7 +55,7 @@ def initialize(cmd, debug, pipe_timeout = 30)
         # wait for the pipe server to signal ready, and fail if no response in 10 seconds
 
         # wait up to 30 seconds in 0.2 second intervals to be able to open the pipe
-        sleep_interval = 0.2
+        sleep_interval = 0.01
```
Iristyle: I'm not sure I understand changing the sleep interval here from 2/10th of a second to 1/100th of a second?

In this loop, we do emit debug output, so this will produce 20x more output while waiting for the pipe to start - which is probably undesirable (even 5x /second is a lot as it stands).

I'm guessing you're doing this as it was proving tricky to get a stalled pipe started up for the sake of tests... I wonder if there's another mechanism we can use to induce that.

RandomNoun7: We were in fact wondering if there was some known reason for picking the original .2 interval value. Yes, we got intermittent passes and fails on the rspec tests with a value as "large" as .2.

michaeltlombardi: You're exactly right. That's something we need to investigate before we merge this in - we were getting passing tests even at `0.2` - maybe we make both the timeout and the sleep duration overwritable?

Iristyle: I just looked at the code again and stand corrected on emitting debug output - I confused this was some other code... so ignore that part of my comment!

Still wondering about practical implications of short(er) sleeps.

michaeltlombardi: Ninja'd by @RandomNoun7. 😛

lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
michaeltlombardi: Also, I think calling `0.times` will just cause an auto failure... so might be a sort-of fix for the sleep duration issue below?

lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
Iristyle: Original 0.2 was fairly random - didn't want to wait for full seconds to elapse while waiting, but didn't want to get into too tight of a sleep / wake loop.  It wasn't chosen very scientifically though, so not sure where we start to get diminishing returns on too short of a sleep time (or even if we do!)

spec/integration/puppet_x/puppetlabs/powershell_manager_spec.rb
```diff
@@ -58,6 +58,13 @@ def create_manager
         expect(first_pid).to eq(second_pid)
       end
 
+      it "should fail if the manger is created with a short timeout" do
+
+        manager.exit
```
Iristyle: You should be able to do this a bit more directly like:

```ruby
expect {
  manager = PowerShellManager.new(manager_args, false, 0.01)
}.to raise_error(RuntimeError, /Failure waiting for PowerShell process \d+ to start pipe server/)
```

I was going to suggest you verify that there's no PID running as well, but that will be hard given how the code is currently written.

lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
michaeltlombardi: Sorry, by "auto failure" I mean it will never assign anything to `@pipe`, so we should always exit as if we timed out, right?

spec/integration/puppet_x/puppetlabs/powershell_manager_spec.rb
michaeltlombardi: Good point. We did verify that there's no process running around behind our backs, but we did that manually outside of ruby.

RandomNoun7: It turned out to be difficult to externally ask the manager you get from PowerShellManager.new(), what it's PID is.

lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
michaeltlombardi: Just verified this: if we set the timeout to something _lower_ than the `sleep_duration, it is always treated as if there is a timeout waiting for the pipe.

lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
RandomNoun7: I'v done a number of tests at this point and I'm not finding a down side to the shorter test interval. You could argue that we risk spiking the processor with context switches, but we risk doing so really only on our own infrastructure machines, and for an exceedingly short period of time.

spec/integration/puppet_x/puppetlabs/powershell_manager_spec.rb
michaeltlombardi: This ended up working:

```ruby
it "should fail if the manger is created with a short timeout" do
  expect {
    manager = PuppetX::PowerShell::PowerShellManager.new(manager_args, false, 0.01)
  }.to raise_error do |e|
    expect(e).to be_a(RuntimeError)
    expected_error = /Failure waiting for PowerShell process (\d+) to start pipe server/
    expect(e.message).to match expected_error
    pid = expected_error.match(e.message)[1].to_i
    expect{Process.kill(0, pid)}.to raise_error(Errno::ESRCH)
  end
end
```

michaeltlombardi:
Updated again with the improved test code, reverted the sleep_interval back to `0.2`

spec/integration/puppet_x/puppetlabs/powershell_manager_spec.rb
```diff
@@ -58,6 +58,18 @@ def create_manager
         expect(first_pid).to eq(second_pid)
       end
 
+      it "should fail if the manger is created with a short timeout" do
+        expect {
+          manager = PuppetX::PowerShell::PowerShellManager.new(manager_args, false, 0.01)
```
Iristyle: Don't need to assign to `manager` here

Iristyle: I think we might want to add at least a little note about the magic of why this works... i.e. anything less than the `sleep_interval` for the manager results in 0 iterations / no waiting:

```
[1] pry(main)> ( 0.01 / 0.2)
=> 0.049999999999999996
[2] pry(main)> ( 0.01 / 0.2).to_int
=> 0
[3] pry(main)> ( 0.01 / 0.2).to_i
=> 0
[4] pry(main)> ( 0.1 / 0.2).to_i
=> 0
[5] pry(main)> ( 0.15 / 0.2).to_i
=> 0
[6] pry(main)> ( 0.2 / 0.2).to_i
=> 1
```

lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
```diff
@@ -9,14 +9,14 @@ module PowerShell
     class PowerShellManager
       @@instances = {}
 
-      def self.instance(cmd, debug = false)
+      def self.instance(cmd, debug = false, pipe_timeout = 30)
```
Iristyle: Do we want defaults on the factory here and on the constructor?

lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
```diff
@@ -55,18 +55,29 @@ def initialize(cmd, debug)
         # wait for the pipe server to signal ready, and fail if no response in 10 seconds
 
         # wait up to 30 seconds in 0.2 second intervals to be able to open the pipe
-        150.times do
+        # If the pipe_timeout is ever specified as less than the sleep interval it will
```
michaeltlombardi: @Iristyle Is this comment sufficient for [this question](https://github.com/puppetlabs/puppetlabs-powershell/pull/202#discussion_r153363126)?

lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
michaeltlombardi: I think the default is only necessary here, not on the constructor, unless I'm missing something / there's a way to call the instructor so it won't inherit from here.

michaeltlombardi: Fixed in the latest push.

spec/integration/puppet_x/puppetlabs/powershell_manager_spec.rb
michaeltlombardi: Fixed in the latest push.

Iristyle:
@michaeltlombardi should be able to squash these commits.

As convention, sometimes we add a note about pairing to the end of the message:

```
Paired-with: Bill Hurt <william.hurt@puppet.com>
```

See https://github.com/puppetlabs/puppet/search?q=paired-with&type=Commits&utf8=%E2%9C%93 for examples in Puppet repo

michaeltlombardi:
@e, @RandomNoun7, 🤞 this should be ready to push through ad-hoc.

PR #213, opened 2018-03-22
Fixing the code example to be correct

Nekototori:

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ For instance, to rename the Guest account:
 ~~~ puppet
 exec { 'rename-guest':
   command   => '$(Get-WMIObject Win32_UserAccount -Filter "Name=\'guest\'").Rename("new-guest")',
-  unless    => 'if (Get-WmiObject Win32_UserAccount -Filter "Name=\'guest\'") { exit 1 }',
+  unless    => 'if (Get-WmiObject Win32_UserAccount -Filter "Name=\'guest\'") { exit 0 }',
   provider  => powershell,
 }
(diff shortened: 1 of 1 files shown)


natemccurdy:
@Nekototori Was the existing logic actually wrong? It doesn't seem like it was.

The `unless` means "only run the command if this check exits non-zero".

And the check was explicitly exiting with a `1` if an account exists named `guest`. With that indicating that we should change the name.

---

Now, an argument could be made that this is quite confusing and we should change the example to use `onlyif`, which tends to be easier to understand.

If we do switch the example to using `onlyif`, then we'd need to change to an explicit `exit 0` and update the description of the example.

Nekototori:
@natemccurdy Ah, that's what I get for getting confused.  Maybe a more straight-forward example (using Powershell 5 commands) and clear up the verbage on the description?
```
exec { 'rename-guest':
  command   => 'Rename-LocalUser -Name Guest -NewName new-guest',
  unless    => 'if (Get-LocalUser Guest) { exit 1 }',
  provider  => powershell,
}
```

natemccurdy:
@Nekototori We definitely should cleanup the example and the description because they don't match up.

I think that PS5 version is good, and I like that it makes the `unless` not too hard to understand.

Just note that if you do decide to go that route, the text before and after the example need to be fixed to be inline with the updated example.

glennsarti:
@Nekototori Here's a blog post I did on exit codes - https://glennsarti.github.io/blog/powershell-puppet-module-exit-codes/

PR #214, opened 2018-03-26
(IMAGES-795) 2008r2 template failing PowerShell module tests

RandomNoun7:
The newest image for Windows 2008r2 is exhibiting different behavior
for loading .NET assemblies than we have seen in the past. The
System.Core assembly is not loaded into the PowerShell host process by
default any longer. This causes the init.ps1 script in the module to
to fail to start a Pipe Server to receive PowerShell scripts to be
executed. This change always attempts to load the System.Core
assembly prior to creating the pipe server object. Explicity loading the
assembly fixes the issue and the newest image becomes usable again. This
fix is also require to fix any moduels that depend on the PowerShell
exec provider working properly.

### DIFF
--- a/lib/puppet_x/templates/powershell/init_ps.ps1
+++ b/lib/puppet_x/templates/powershell/init_ps.ps1
@@ -748,6 +748,8 @@ function Start-PipeServer
     $Encoding
   )
 
+  Add-Type -AssemblyName "System.Core, Version=3.5.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089"
+
   # this does not require versioning in the payload as client / server are tightly coupled
   $server = New-Object System.IO.Pipes.NamedPipeServerStream($CommandChannelPipeName,
     [System.IO.Pipes.PipeDirection]::InOut)


RandomNoun7:
To test this pull request, you will need to test on vmpooler. Since the 2008r2 template in CINext has been rolled forward to temporarily deal with this issue, a green Jenkins run means nothing.

Clone down the repo and my branch. Do a `bundle exec rake beaker` run against master and watch it fail against the vmpooler images. Then checkout this branch, re run the tests, and watch the tests go green.

glennsarti:
Confirmed.

Iristyle:
Will leave the same comment as I left on the DSC PR - we shouldn't be using `IMAGES` tickets for modules.

lib/puppet_x/templates/powershell/init_ps.ps1
```diff
@@ -748,6 +748,8 @@ function Start-PipeServer
     $Encoding
   )
 
+  Add-Type -AssemblyName "System.Core, Version=3.5.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089"
```
Iristyle: What about for systems that don't have the `3.5.0.0` version of `System.Core`?

Iristyle: I suspect PowerShell has binding redirects setup, but I have no confirmation of this yet ... On our 2016 templates, there is no `System.Core 3.5.0.0` - but `System.Core 4.0.30319` is already loaded into the PowerShell process AppDomain. The call to `Add-Type` with the explicit version does not fail. However, its unclear if that's due to `System.Core` already being loaded / the call being ignored or if its due to first redirecting the `3.5.0.0` request to a `4.0.30319` request and then determining the assembly is already loaded.

More info on binding redirects:
https://docs.microsoft.com/en-us/dotnet/framework/configure-apps/redirect-assembly-versions

I'm digging in with the fusion log viewer to see if I can get any more details:
https://docs.microsoft.com/en-us/dotnet/framework/tools/fuslogvw-exe-assembly-binding-log-viewer

Iristyle: As a quick follow-up, I've confirmed behavior using the fusion log viewer. When requesting `System.Core 3.5.0.0`, fusion redirects that request to `System.Core, Version=4.0.0.0`. Since that assembly is already loaded, `Add-Type` succeeds:

```
*** Assembly Binder Log Entry  (3/27/2018 @ 3:35:39 PM) ***

The operation was successful.
Bind result: hr = 0x0. The operation completed successfully.

Assembly manager loaded from:  C:\Windows\Microsoft.NET\Framework64\v4.0.30319\clr.dll
Running under executable  C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe
--- A detailed error log follows. 

=== Pre-bind state information ===
LOG: DisplayName = System.Core, Version=3.5.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089
 (Fully-specified)
LOG: Appbase = file:///C:/Windows/System32/WindowsPowerShell/v1.0/
LOG: Initial PrivatePath = NULL
LOG: Dynamic Base = NULL
LOG: Cache Base = NULL
LOG: AppName = powershell.exe
Calling assembly : (Unknown).
===
LOG: This bind starts in default load context.
LOG: No application configuration file found.
LOG: Using host configuration file: 
LOG: Using machine configuration file from C:\Windows\Microsoft.NET\Framework64\v4.0.30319\config\machine.config.
LOG: Version redirect found in framework config: 3.5.0.0 redirected to 4.0.0.0.
LOG: Post-policy reference: System.Core, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089
LOG: Binding succeeds. Returns assembly from C:\Windows\Microsoft.Net\assembly\GAC_MSIL\System.Core\v4.0_4.0.0.0__b77a5c561934e089\System.Core.dll.
LOG: Assembly is loaded in default load context.
```

I suspect that without a redirect in place the load will instead fail, so I think the best bet here is probably to stick with the simpler `Add-Type -AssemblyName 'System.Core'` so that for machines that may not perform redirects back to 3.5.0.0, the code will still work (which might be the case in a PS6 or cross-platform scenario). We're less concerned with the version of the assembly, and more concerned with gaining access to a few types within it.

RandomNoun7:
Reverting this PR to deal to remove the specific version in the `Add-Type` call.

PR #223, opened 2018-04-20
(MODULES-7033) Kill test unreliable

RandomNoun7:
This change fixes a spec test that was giving unreliable results and erroneously
failing CI pipelines.


lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
```diff
@@ -75,8 +75,10 @@ def initialize(cmd, debug, pipe_timeout)
           # clutter behind, so explicitly close those too.
           @stdout.close if !@stdout.closed?
           @stderr.close if !@stderr.closed?
-          Process.kill("KILL", @ps_process[:pid]) if @ps_process.alive?
-          raise "Failure waiting for PowerShell process #{@ps_process[:pid]} to start pipe server"
+          kill_called = Process.kill("KILL", @ps_process[:pid]) if @ps_process.alive?
```
Iristyle: So I don't think the return value here means kill called... see https://github.com/ruby/ruby/blob/trunk/signal.c#L413-L532. In particular, the return is `INT2FIX(i-1);`, which Ruby docs don't even document. http://www.rubydoc.info/stdlib/core/Process.kill

spec/integration/puppet_x/puppetlabs/powershell_manager_spec.rb
```diff
@@ -63,10 +63,8 @@ def create_manager()
           PuppetX::PowerShell::PowerShellManager.new(manager_args, false, 0.01)
         }.to raise_error do |e|
           expect(e).to be_a(RuntimeError)
-          expected_error = /Failure waiting for PowerShell process (\d+) to start pipe server/
+          expected_error = /Failure waiting for PowerShell process \d+ to start pipe server. Process killed/
           expect(e.message).to match expected_error
-          pid = expected_error.match(e.message)[1].to_i
```
Iristyle: So this assertion originally verified that the process was no longer running.

RandomNoun7:
It looks like the race condition was calling Process.kill inside the code are testing, and then calling Process.kill again in the rspec code, and not always allowing enough time for the process to die before we try to verify that it's dead. Adding a second into the test before calling kill again seems to allow the required time for the process to finish dying. @Iristyle This is ready for review and merge.

spec/integration/puppet_x/puppetlabs/powershell_manager_spec.rb
```diff
@@ -66,6 +66,7 @@ def create_manager()
           expected_error = /Failure waiting for PowerShell process (\d+) to start pipe server/
           expect(e.message).to match expected_error
           pid = expected_error.match(e.message)[1].to_i
+          sleep(1)
```
Iristyle: It would be nice to have a comment here on how we're avoiding a race by calling sleep to wait for the terminated process to complete it's cleanup. As we've discovered, even though `KILL` was called, the PID may not yet be removed from the process table just yet.

RandomNoun7:
@Iristyle Comment added.

PR #238, opened 2018-12-13
(MODULES-8355) Add pwsh provider

glennsarti:
PowerShell 6 now uses a different binary and has different installation locations
compared to Windows PowerShell.  This commit adds a new provider called pwsh
which will execute scripts but using PowerShell 6 (pwsh).  Later commits will add
advanced features, but this commit adds a basic provider with tests that works
only on Windows, with PowerShell 6 in the default installation location.

### DIFF
--- a/lib/puppet/provider/exec/pwsh.rb
+++ b/lib/puppet/provider/exec/pwsh.rb
@@ -1,0 +1,70 @@
+require 'puppet/provider/exec'
+
+Puppet::Type.type(:exec).provide :pwsh, :parent => Puppet::Provider::Exec do
+  confine :operatingsystem  => [ :windows ]
+
+  commands :pwsh =>
+    if File.exists?("#{ENV['ProgramFiles']}\\PowerShell\\6\\pwsh.exe")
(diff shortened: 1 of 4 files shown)


lib/puppet/provider/exec/pwsh.rb
```diff
@@ -0,0 +1,63 @@
...
+  commands :pwsh => 'pwsh.exe'
+
+  desc <<-EOT
+    Executes PowerShell Core commands. One of the `onlyif`, `unless`, or `creates`
+    parameters should be specified to ensure the command is idempotent.
+
+    Example:
+        # Rename the Guest account
+        exec { 'rename-guest':
+          command   => '$(Get-WMIObject Win32_UserAccount -Filter "Name=\'guest\'").Rename("new-guest")',
```
jpogran: This example needs to be changed. Use `Get-CimInstance` instead of `Get-WMIObject` as WMI is not present in pwsh.

glennsarti: Fixed

glennsarti:
Note I had done manual beaker runs with a `powershell6` role on Windows and they pass successfully.

PR #240, opened 2019-01-09
(MODULES-8357) Add cross platform support for pwsh provider

glennsarti:
Previously the pwsh provider only functioned on Windows operating systems.

This commit;
* Modifies the provider to not be confined to Windows operating systems
* Modifies the spec tests to be able to run on non-Windows platforms
* Modifies the acceptance tests to be able to run on non-Windows platforms

### DIFF
--- a/lib/puppet/provider/exec/pwsh.rb
+++ b/lib/puppet/provider/exec/pwsh.rb
@@ -1,10 +1,10 @@
 require 'puppet/provider/exec'
 
 Puppet::Type.type(:exec).provide :pwsh, :parent => Puppet::Provider::Exec do
-  confine :operatingsystem  => [ :windows ]
-
   commands :pwsh =>
-    if File.exists?("#{ENV['ProgramFiles']}\\PowerShell\\6\\pwsh.exe")
+    if !Puppet::Util::Platform.windows?
+      'pwsh'
+    elsif File.exists?("#{ENV['ProgramFiles']}\\PowerShell\\6\\pwsh.exe")
       "#{ENV['ProgramFiles']}\\PowerShell\\6\\pwsh.exe"
     elsif File.exists?("#{ENV['ProgramFiles(x86)']}\\PowerShell\\6\\pwsh.exe")
       "#{ENV['ProgramFiles(x86)']}\\PowerShell\\6\\pwsh.exe"
@@ -35,7 +35,11 @@ Puppet::Type.type(:exec).provide :pwsh, :parent => Puppet::Provider::Exec do
       # we redirect powershell's stdin to read from the file. Current
       # versions of Windows use per-user temp directories with strong
       # permissions, but I'd rather not make (poor) assumptions.
-      return super("cmd.exe /c \"\"#{native_path(command(:pwsh))}\" #{legacy_args} -Command - < \"#{native_path}\"\"", check)
+      if Puppet::Util::Platform.windows?
+        return super("cmd.exe /c \"\"#{native_path(command(:pwsh))}\" #{legacy_args} -Command - < \"#{native_path}\"\"", check)
+      else
+        return super("/bin/sh -c \"#{native_path(command(:pwsh))} #{legacy_args} -Command - < #{native_path}\"", check)
+      end
     end
   end
 
--- a/spec/acceptance/exec_pwsh_spec.rb
+++ b/spec/acceptance/exec_pwsh_spec.rb
@@ -1,21 +1,14 @@
 require 'spec_helper_acceptance'
 
-describe 'powershell provider:' do
+def windows_platform?(host)
+  !((host.platform =~ /^windows.*$/).nil?)
+end
 
-  powershell6_agents = hosts_as('powershell6')
-  posix6_agents      = powershell6_agents.select { |a| a.platform =~ /^(?!windows).*$/ }
-  windows6_agents    = powershell6_agents.select { |a| a.platform =~ /^windows.*$/ }
+powershell6_agents = hosts_as('powershell6')
+posix6_agents      = powershell6_agents.select { |a| !windows_platform?(a) }
+windows6_agents    = powershell6_agents.select { |a| windows_platform?(a) }
 
-  # Due to https://github.com/PowerShell/PowerShell/issues/1794 the HOME directory must be passed in the environment explicitly
-  # In this case, it just needs a HOME that has a valid directory, no files get stored there
-  # HOME is not used on Windows so it is safe to apply hosts, no matter its platform
-  let (:ps_environment) { "environment => ['HOME=/tmp']," }
-  ps_environment = "environment => ['HOME=/tmp'],"
-
-  def windows_platform?(host)
-    !((host.platform =~ /^windows.*$/).nil?)
-  end
-
+describe 'pwsh provider:', :if => powershell6_agents.count > 0 do
   def platform_string(host, windows, posix)
     if windows_platform?(host)
       windows
@@ -26,7 +19,6 @@ describe 'powershell provider:' do
 
   shared_examples 'should fail' do |manifest, error_check|
     it 'should throw an error' do
-      # Due to https://tickets.puppetlabs.com/browse/QA-3461 each host must be done one at a time
       powershell6_agents.each do |hut|
         result = execute_manifest_on(hut, manifest, :expect_failures => true)
         unless error_check.nil?
@@ -38,22 +30,19 @@ describe 'powershell provider:' do
 
   shared_examples 'apply success' do |manifest|
     it 'should succeed' do
-      # Due to https://tickets.puppetlabs.com/browse/QA-3461 each host must be done one at a time
-      powershell6_agents.each { |hut| execute_manifest_on(hut, manifest, :catch_failures => true) }
+      execute_manifest_on(powershell6_agents, manifest, :catch_failures => true)
     end
   end
 
-  shared_examples 'standard exec' do |powershell_cmd, host_list|
+  shared_examples 'standard exec' do |powershell_cmd, hosts|
     padmin = <<-MANIFEST
       exec{'no fail test':
         command  => '#{powershell_cmd}',
-        #{ps_environment}
         provider => pwsh,
       }
     MANIFEST
     it 'should not fail' do
-      # Due to https://tickets.puppetlabs.com/browse/QA-3461 each host must be done one at a time
-      host_list.each { |hut| execute_manifest_on(hut, padmin, :catch_failures => true) }
+      execute_manifest_on(hosts, padmin, :catch_failures => true)
     end
   end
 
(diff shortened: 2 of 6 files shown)


lib/puppet/provider/exec/pwsh.rb
```diff
@@ -26,16 +26,28 @@
...
+      end
+    else
+      write_script(command) do |native_path|
+        # Ideally, we could keep a handle open on the temp file in this
+        # process (to prevent TOCTOU attacks), and execute powershell
+        # with -File <path>. But powershell complains that it can't open
+        # the file for exclusive access. If we close the handle, then an
+        # attacker could modify the file before we invoke powershell. So
+        # we redirect powershell's stdin to read from the file.
+        return super("sh -c \"#{native_path(command(:pwsh))} #{posix_args} -Command - < #{native_path}\"", check)
```
Iristyle: I'm not sure that `sh` is going to be universally correct across all platforms - so this may be something we want to verify in some manual tests first?

lib/puppet/provider/exec/pwsh.rb
```diff
@@ -67,4 +79,10 @@ def native_path(path)
   def legacy_args
     '-NoProfile -NonInteractive -NoLogo -ExecutionPolicy Bypass'
   end
+
+  def posix_args
+    # Note - using -ExecutionPolicy causes PowerShell to abort
```
Iristyle: Is this still an issue? The linked GitHub issue is closed as fixed in 6 Alpha 18.

spec/acceptance/exec_pwsh_spec.rb
```diff
@@ -438,6 +438,18 @@ def platform_string(host, windows, posix)
       it { should be_file }
       its(:content) { should match /puppet/ }
     end
+
+    describe file('/services.txt'), :if => posix6_agents.count > 0 do
```
Iristyle: Can we just collapse this into one platform-agnostic test rather than 2 nearly identical tests?

spec/acceptance/exec_pwsh_spec.rb
```diff
@@ -482,6 +494,27 @@ def platform_string(host, windows, posix)
...
+      }
+      exec{"TestPowershellPS1":
+        command   => '/external-script.ps1',
+        #{ps_environment}
+        provider  => powershell,
+        require   => File['/external-script.ps1']
+      }
+      MANIFEST
+
+      describe file('/tmp/commands.csv') do
```
Iristyle: Is there a Beaker helper we can use for a cross-platform TEMP?

Some mods here, would allow us to again share the tests cross platform.

spec/acceptance/exec_pwsh_spec.rb
```diff
@@ -507,6 +540,31 @@ def platform_string(host, windows, posix)
         its(:content) { should match /svchost/ }
       end
     end
+
+    context 'on POSIX platforms', :if => posix6_agents.count > 0 do
```
Iristyle: Another one where I'd rather see us share a test and make it universal if possible.

spec/unit/provider/exec/pwsh_spec.rb
```diff
@@ -114,6 +131,45 @@
           expect(status.exitstatus).to eq(1)
         end
       end
+
+      context "on non-Windows", :if => !Puppet.features.microsoft_windows? do
+        # The usage of uname is a little fragile however there is basically nothing
+        # which is universal across all Linux/Unix/Mac distributions; Unlike Well Known SIDS in Windows
+        # The closest is the presence of the uname command and its generic text output
+        let(:command) { '& uname' }
```
Iristyle: Is it important to the test hat we shell out to a command like this?

lib/puppet/provider/exec/pwsh.rb
glennsarti: Confirmed not an issue.  will remove.

lib/puppet/provider/exec/pwsh.rb
glennsarti: should I use `/bin/sh` as per https://github.com/puppetlabs/puppet/blob/master/lib/puppet/provider/exec/shell.rb#L19 or perhaps get rid of the shell call entirely.

spec/unit/provider/exec/pwsh_spec.rb
glennsarti: In this case we're testing stdin/stdout streams are being passed from external programs back to us in the provider.  This is really only important on the PowerShell Manager side of things, but still "useful" for the script based invocations.

lib/puppet/provider/exec/pwsh.rb
glennsarti: Fixed

spec/acceptance/exec_pwsh_spec.rb
glennsarti: Done.

spec/acceptance/exec_pwsh_spec.rb
glennsarti: Done.

lib/puppet/provider/exec/pwsh.rb
Iristyle: Yeah, `/bin/sh` seems sensible if we've already vetted that across other platforms and have standardized on that within `exec` already

glennsarti: Fixed

spec/acceptance/exec_pwsh_spec.rb
glennsarti: With my new changes this is now redundant.

spec/acceptance/exec_pwsh_spec.rb
```diff
@@ -340,111 +298,99 @@ def platform_string(host, windows, posix)
...
-      let(:file_path) { 'C:/services.txt' }
+    win_file = 'C:/services.txt'
+    posix_file = '/services.txt'
 
-      it 'should apply the manifest' do
-        # Due to https://tickets.puppetlabs.com/browse/QA-3461 each host must be done one at a time
-        windows6_agents.each { |hut| execute_manifest_on(hut, p2, :catch_failures => true) }
-      end
+    powershell6_agents.each do |node|
+      describe file(windows_platform?(node) ? win_file : posix_file), :node => node do
```
Iristyle: Much better IMHO

spec/acceptance/exec_pwsh_spec.rb
```diff
@@ -456,55 +402,76 @@ def platform_string(host, windows, posix)
...
+    win_file = 'c:/temp/commands.csv'
+    posix_file = '/tmp/commands.csv'
+
+    powershell6_agents.each do |node|
+      describe file(windows_platform?(node) ? win_file : posix_file), :node => node do
+        let(:external_script) { windows_platform?(node) ? 'c:/external-script.ps1' : '/external-script.ps1' }
+        let(:external_fixture) { "files/get-command-#{platform_string(node, 'win', 'posix')}.ps1" }
+
+        it 'should apply the manifest' do
+          #require 'pry'; binding.pry
```
Iristyle: Straggler here @glennsarti

spec/acceptance/exec_pwsh_spec.rb
```diff
@@ -456,55 +402,76 @@ def platform_string(host, windows, posix)
...
+    }
+    MANIFEST
+    }
+
+    win_file = 'c:/temp/params.csv'
+    posix_file = '/tmp/params.csv'
+
+    powershell6_agents.each do |node|
+      describe file(windows_platform?(node) ? win_file : posix_file), :node => node do
+        let(:external_script) { windows_platform?(node) ? 'C:\\param_script.ps1' : '/param_script.ps1' }
```
Iristyle: Did you try using just `/param_script.ps1` across all platforms?

spec/acceptance/exec_pwsh_spec.rb
glennsarti: Doh.

spec/acceptance/exec_pwsh_spec.rb
```diff
@@ -456,55 +402,76 @@ def platform_string(host, windows, posix)
...
+        let(:commandName) { 'Export-Csv' }
+
+
+        it 'should apply the manifest' do
+          execute_manifest_on(node, manifest, :catch_failures => true)
+        end
+
         it { should be_file }
-        its(:content) { should match /svchost/ }
+        its(:content) { should match /Export-Csv/ }
```
Iristyle: Should we reuse `commandName` here?

spec/acceptance/exec_pwsh_spec.rb
glennsarti: No I didn't. but  I given I had to do it differently per platform anyway, it seemed the right thing to do

spec/acceptance/exec_pwsh_spec.rb
glennsarti: We could yeah. Will do.

spec/unit/provider/exec/pwsh_spec.rb
```diff
@@ -39,80 +33,84 @@
...
 
     context "actual runs" do
-      context "on Windows", :if => Puppet.features.microsoft_windows? do
-        it "returns the output and status" do
-          output, status = provider.run(command)
+      # The usage of uname is a little fragile however there is basically nothing
+      # which is universal across all Linux/Unix/Mac distributions; Unlike Well Known SIDS in Windows
+      # The closest is the presence of the uname command and its generic text output
+      let(:command) { Puppet.features.microsoft_windows? ?
+        '$(Get-CIMInstance Win32_Account -Filter "SID=\'S-1-5-18\'") | Format-List' :
```
Iristyle: Not sure if it's worth doing... but can't we just use `Get-CimInstance` on all platforms (with a different type I suppose)?

glennsarti: Not sure.  Will need to look into it more.

Iristyle: Ah no... looks like CIM classes aren't part of PowerShell itself - so might not be anything there to choose from anyhow.

spec/acceptance/exec_pwsh_spec.rb
glennsarti: Fixed

spec/acceptance/exec_pwsh_spec.rb
glennsarti: Fixed

Iristyle:
Oh oh... so I think what you're saying here is:

* Windows had 29 tests (?), now 60 (8:11)
    * Not sure if "files took 5 minutes 11 seconds to load" is something we should worry about... 
* Ubuntu had 27 tests (2:42), now 56 (5:08)

glennsarti:
`files took 5 minutes 11 seconds to load" is something we should worry about...`   <--- Australian Internet to PDX based VMs :-)

PR #241, opened 2019-01-21
(MODULES-8356) Search for pwsh binary

glennsarti:
~~Builds on PR #245~~

---

Previously the pwsh provider only looked in the system path and some
predefined paths on Windows.  However PowerShell Core can be installed
anywhere as it is now a tarball/zipfile installation.  This means a
user should be able to arbitrarily define a location for the provider
to use.

This commit removes the :pwsh command because command resolution
can not use a custom path when searching. This is normally used to see
if a provider is applicable.  It is then replaced by a method called
get_pwsh_command which uses a copy of the Puppet `which` command which
can use custom search paths when locating binaries.

Note we can not extend or update the Puppet version of this code as
the which method is private. In this case we copy the code verbatim
and then modify it.

This commit also will now fail a catalog apply if pwsh can not be
located. Previously it would error that no suitable provider could
be found.

This commit also updates the READMe and metadata.json to reflect the
operating systems that the pwsh provider will run on.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,13 +20,25 @@ This module adds a new exec provider capable of executing PowerShell commands.
 
 ## Module Description
 
-Puppet provides a built-in `exec` type that is capable of executing commands. This module adds a `powershell` provider to the `exec` type,  which enables `exec` parameters, listed below. This module is particularly helpful if you need to run PowerShell commands but don't know the details about how PowerShell is executed, because you can run PowerShell commands in Puppet without the module.
+Puppet provides a built-in `exec` type that is capable of executing commands. This module adds a `powershell` and `pwsh` provider to the `exec` type,  which enables `exec` parameters, listed below. This module is particularly helpful if you need to run PowerShell commands but don't know how PowerShell is executed, because you can run PowerShell commands in Puppet without the module.
 
 ## Setup
 
 ### Requirements
 
-This module requires PowerShell to be installed and the `powershell.exe` to be available in the system PATH.
+The `powershell` provider requires [Windows PowerShell](https://docs.microsoft.com/en-us/powershell/scripting/install/installing-powershell?view=powershell-5.1) and have `powershell.exe` available in the system PATH. Note that most Windows operating systems already have Windows PowerShell installed.
+
+The `pwsh` provider requires you install [PowerShell Core](https://docs.microsoft.com/en-us/powershell/scripting/install/installing-powershell?view=powershell-6) and make `pwsh` available either in the system PATH or specified in the `path` parameter.
+
+For example, when PowerShell Core is installed in `/usr/alice/pscore` the following manifest is needed:
+
+~~~ puppet
+exec { 'RESOURCENAME':
+  ...
+  path     => '/usr/alice/pscore',
+  provider => pwsh,
+}
+~~~
 
 ### Beginning with powershell
 
@@ -41,7 +53,7 @@ exec { 'RESOURCENAME':
 
 ## Usage
 
-When using `exec` resources with the `powershell` provider, the `command` parameter must be single-quoted to prevent Puppet from interpolating `$(..)`.
+When using `exec` resources with the `powershell` or `pwsh` provider, the `command` parameter must be single-quoted to prevent Puppet from interpolating `$(..)`.
 
 For instance, to rename the Guest account:
 
@@ -129,7 +141,9 @@ However, to produce output from a script, use the `Write-` prefixed cmdlets such
 
 #### Provider
 
-* powershell: Adapts the Puppet `exec` resource to run PowerShell commands.
+* powershell: Adapts the Puppet `exec` resource to run [Windows PowerShell](https://docs.microsoft.com/en-us/powershell/scripting/install/installing-powershell?view=powershell-5.1) commands.
+
+* pwsh: Adapts the Puppet `exec` resource to run [PowerShell Core](https://docs.microsoft.com/en-us/powershell/scripting/install/installing-powershell?view=powershell-6) commands.
 
 #### Parameters
 
@@ -163,6 +177,8 @@ Runs the exec only if the command returns 0. Valid options: String. Default: Und
 
 Specifies the search path used for command execution. Valid options: String of the path, an array, or a semicolon-separated list. Default: Undefined.
 
+The `pwsh` provider can also use the path to find the pwsh executable.
+
 ##### `refresh`
 
 Refreshes the command. Valid options: String. Default: Undefined.
@@ -193,11 +209,29 @@ Runs the `exec`, unless the command returns 0. Valid options: String. Default: U
 
 ## Limitations
 
-* Only supported on Windows Server 2008 and above, and Windows 7 and above.
+* The `powershell` provider is only supported on Windows Server 2008 and above, and Windows 7 and above.
 
-* Experimental support added for Ubuntu 16.04, Ubuntu 14.0.4 and, CentOS 7. Note that this module will not install PowerShell on these platforms. For further information see the [Linux installation instructions](https://docs.microsoft.com/en-us/powershell/scripting/setup/installing-powershell-core-on-linux).
+* The `pwsh` provider is supported on:
 
-* Only supported on Powershell 2.0 and above.
+  * CentOS 7
+
+  * Debian 8.7+, Debian 9
+
+  * Fedora 27, 28
+
+  * MacOS 10.12+
+
+  * Red Hat Enterprise Linux 7
+
+  * Ubuntu 14.04, 16.0.4 and 18.04
+
+  * Windows Desktop 7 and above
+
+  * Windows Server 2008 R2 and above
+
+  Note that this module will not install PowerShell on these platforms. For further information see the [Linux installation instructions](https://docs.microsoft.com/en-us/powershell/scripting/setup/installing-powershell-core-on-linux).
+
+* Only supported on Windows PowerShell 2.0 and above, and PowerShell Core 6.1 and above.
 
 * When using here-strings in inline or templated scripts executed by this module, you must use the double-quote style syntax that begins with `@"` and ends with `"@`. The single-quote syntax that begins with `@'` and ends with `'@` is not supported.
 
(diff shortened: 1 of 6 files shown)


glennsarti:
Blocked by #242

README.md
```diff
@@ -20,13 +20,25 @@ This module adds a new exec provider capable of executing PowerShell commands.
 
 ## Module Description
 
-Puppet provides a built-in `exec` type that is capable of executing commands. This module adds a `powershell` provider to the `exec` type,  which enables `exec` parameters, listed below. This module is particularly helpful if you need to run PowerShell commands but don't know the details about how PowerShell is executed, because you can run PowerShell commands in Puppet without the module.
+Puppet provides a built-in `exec` type that is capable of executing commands. This module adds a `powershell` and `pwsh` provider to the `exec` type,  which enables `exec` parameters, listed below. This module is particularly helpful if you need to run PowerShell commands but don't know the details about how PowerShell is executed, because you can run PowerShell commands in Puppet without the module.
```
clairecadman: Suggest omitting "the details about":

"....if you need to run PowerShell commands but don't know how PowerShell is executed..."

README.md
```diff
@@ -20,13 +20,25 @@ This module adds a new exec provider capable of executing PowerShell commands.
...
 
 ## Setup
 
 ### Requirements
 
-This module requires PowerShell to be installed and the `powershell.exe` to be available in the system PATH.
+The `powershell` provider requires [Windows PowerShell](https://docs.microsoft.com/en-us/powershell/scripting/install/installing-powershell?view=powershell-5.1) to be installed and the `powershell.exe` to be available in the system PATH.
```
clairecadman: Suggested reword:

"The `powershell` provider requires you install [Windows PowerShell] and have the `powershell.exe` available in the system PATH."

README.md
```diff
@@ -20,13 +20,25 @@ This module adds a new exec provider capable of executing PowerShell commands.
...
+
+The `pwsh` provider requires [PowerShell Core](https://docs.microsoft.com/en-us/powershell/scripting/install/installing-powershell?view=powershell-6) to be installed and the `pwsh` to be available either in the system PATH or specified in the `path` parameter.
```
clairecadman: Same suggestion as above:

"The `pwsh` provider requires you install [PowerShell Core] and make `pwsh` available either in the system PATH or specified in the `path` parameter."

README.md
```diff
@@ -20,13 +20,25 @@ This module adds a new exec provider capable of executing PowerShell commands.
...
+
+For example, when PowerShell Core is installed in `/usr/alice/pscore` the following manifest is needed;
```
clairecadman: : instead of ;

README.md
```diff
@@ -193,11 +209,29 @@ Runs the `exec`, unless the command returns 0. Valid options: String. Default: U
 
 ## Limitations
 
-* Only supported on Windows Server 2008 and above, and Windows 7 and above.
+* The `powershell` provider is only supported on Windows Server 2008 and above, and Windows 7 and above.
+
+* The `pwsh` provider is supported on
```
clairecadman: Colon after "The `pwsh` provider is supported on"

spec/unit/provider/exec/pwsh_spec.rb
```diff
@@ -68,6 +56,28 @@
 
         provider.run_spec_override(command)
       end
+
+      context 'when specifying a path' do
+        let(:path) { Puppet::Util::Platform.windows? ? 'C:/pwsh-test' : '/pwsh-test' }
```
ThoughtCrhyme: So, the problem I was seeing on my machine (and a centos vmpooler machine) is that .[travis.yaml](https://github.com/puppetlabs/puppetlabs-powershell/blob/master/.travis.yml#L10) installs powershell on the SUT. However the machines I was using did not have powershell installed and thus the code at provider/exec/pwsh.rb:19 is throwing "pwsh could not be found".

This dependency should be stubbed out. We should not need to actually install pwsh to complete the unit tests.

glennsarti:
@Iristyle Yeah, revert was to keep future diffs clean.

spec/unit/provider/exec/pwsh_spec.rb
glennsarti: Right ok.... I've kind of solved this in two parts;

Firstly, those tests require _actually_ calling powershell/pwsh so I've moved those particular tests to `/spec/integration`. This is more admin than fixing though.  I did fix the unit tests to properly stub out any filesystem calls for pwsh so the unit tests will run whether or not pwsh is available.

What I also did is add a guard on those tests to detect if `pwsh` is available and skip them if not.  I'm not 100% happy that I am, but at least it will obvious in the rspec logs e.g.

```
Puppet::Type::Exec::ProviderPwsh
  #run
    returns the output and status (PENDING: Could not locate pwsh binary)
    returns true if the `onlyif` check command succeeds (PENDING: Could not locate pwsh binary)
    returns false if the `unless` check command succeeds (PENDING: Could not locate pwsh binary)
    runs commands properly that output to multiple streams (PENDING: Could not locate pwsh binary)

Pending: (Failures listed here are expected and do not affect your suite's status)

  1) Puppet::Type::Exec::ProviderPwsh#run returns the output and status
     # Could not locate pwsh binary
     # ./spec/integration/provider/exec/pwsh_spec.rb:33

  2) Puppet::Type::Exec::ProviderPwsh#run returns true if the `onlyif` check command succeeds
     # Could not locate pwsh binary
     # ./spec/integration/provider/exec/pwsh_spec.rb:41

  3) Puppet::Type::Exec::ProviderPwsh#run returns false if the `unless` check command succeeds
     # Could not locate pwsh binary
     # ./spec/integration/provider/exec/pwsh_spec.rb:48

  4) Puppet::Type::Exec::ProviderPwsh#run runs commands properly that output to multiple streams
     # Could not locate pwsh binary
     # ./spec/integration/provider/exec/pwsh_spec.rb:55
```

Secondly I removed the non-Windows unit tests from the `powershell` provider.  Ideally I wanted to do this as part of MODULES-8359, but after consideration I didn't think I should bother making changes to make the tests pass, to only delete all that effort in 8359 later on.  So instead I just removed the unit tests only.

glennsarti: Oh to answer, how did I test this locally, I just renamed `C:\Program Files\PowerShell\6\pwsh.exe` to `pwsh-hidden.exe` so the provider will fail.

glennsarti:
@Iristyle I've moved the PDK template changes to a different PR and then build upon that.  Too many changes polluting this PR.

glennsarti:
Rebased onto master.

lib/puppet/provider/exec/pwsh.rb
```diff
@@ -51,6 +42,71 @@ def validatecmd(command)
...
+    pwsh = "#{ENV['ProgramFiles']}\\PowerShell\\6\\pwsh.exe"
+    return pwsh if File.exists?(pwsh)
+    pwsh = "#{ENV['ProgramFiles(x86)']}\\PowerShell\\6\\pwsh.exe"
+    File.exists?(pwsh) ? pwsh : nil
+  end
+
+  # Based on which command from https://github.com/puppetlabs/puppet/blob/1c14d0a9fdfc31933603e571b616b6cd675e6b71/lib/puppet/util.rb#L241-L286
+  # Resolve a path for an executable to the absolute path. This tries to behave
+  # in the same manner as the unix `which` command and uses the `PATH`
+  # environment variable.
```
Iristyle: I don't like that we're reimplementing code that already exists in Puppet. Couldn't you just wrap the existing Puppet call using withenv? https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util.rb#L119-L146

Usage in the existing `exec` is something like this - 

```ruby
if resource[:path]
  Puppet::Util.withenv( {'PATH' => resource[:path].join(File::PATH_SEPARATOR)}, :windows) do
    return if which(exe)
  end
end
```

https://github.com/puppetlabs/puppet/blob/ba4d1a1aba0095d3c70b98fea5c67434a4876a61/lib/puppet/provider/exec/windows.rb#L47-L51

lib/puppet/provider/exec/pwsh.rb
```diff
@@ -26,6 +15,8 @@
   EOT
 
   def run(command, check = false)
+    pwsh = get_pwsh_command
```
Iristyle: Are you saying the `:path` attribute cannot take effect when resolving this `commands` block? I thought there were ways to defer execution here using a lambda, but will have to double check that for sure.

lib/puppet/provider/exec/pwsh.rb
glennsarti: I don't like it either but that method is marked `private` so I shouldn't be calling it.  I could monkey patch a public method but that seems even worse.

Given that it's a "fairly" static method I decided to copy and modify instead.

lib/puppet/provider/exec/pwsh.rb
glennsarti: TLDR Yes.  When `commands` is being evaluated it had no access to the resource so there was no way to get the path.

lib/puppet/provider/exec/pwsh.rb
Iristyle: Both `withenv` at https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util.rb#L119-L135 and `which` at https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util.rb#L241-L286 are module functions exposed on `Puppet::Util`.

Not sure I understand?

glennsarti: You are right...apologies

glennsarti: With your code example above.  That will work, it just seems heavy handed to change the search path temporarily instead of just modifying the which command.

Also while _probably_ not a problem for MRI ruby, it wouldn't be thread safe for real threads.

Iristyle: This is the pattern we've established in Puppet itself, so it's preferable to stay consistent... thread safety isn't a concern here.

PR #242, opened 2019-01-24
(maint) Fix module installing in master-agent scenario

glennsarti:
Previously the powershell module was only being installed onto the
default host.  However this meant the module was only being installed
onto the agent, not the server, which then caused issues during tests
because the module content was not 'file syncd' from the master and
tests were failing as it couldn't find the powershell provider.

This commit uses the module install helper from beaker to install on
all hosts.


glennsarti:
Running through adhoc...

glennsarti:
Passed adhoc at https://jenkins-master-prod-1.delivery.puppetlabs.net/view/modules/view/windows/view/_adhoc/view/powershell/job/forge-windows_puppetlabs-powershell_init-cinext_nightly-skippable_adhoc/3/

PR #244, opened 2019-01-25
(maint) Add HTML anchor tag

clairecadman:
The anchor tag helps fix broken links — If the header text changes (gets localized) the link will brake.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,20 +14,25 @@
 5. [Limitations - OS compatibility, etc.](#limitations)
 6. [Development - Guide for contributing to the module](#development)
 
+<a id="overview"></a>
 ## Overview
 
 This module adds a new exec provider capable of executing PowerShell commands.
 
+<a id="module-description"></a>
 ## Module Description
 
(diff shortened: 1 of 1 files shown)


joshcooper:
Unfortunately we discovered that anchors don't work, because the forge strips them when rendering the README. Not sure what next steps are @ehom and @scotje?

scotje:
As far as I can tell in my testing, the method utilized here:

https://github.com/puppetlabs/puppetlabs-yumrepo_core/compare/test/master

Seems to work fine in the browsers I tested with.

ehom:
@joshcooper @clairecadman 
we are currently looking into migrating all markdown files to our documentation locaiization platform.

if we take out the HTML anchors, the links to the page anchors will not work since Transifex does not expose the links to the anchors in the localization editor. If we are okay with that, will can remove the HTML anchors.

clairecadman:
@ehom Okay! Should I remove them from all of them or just this one?

PR #247, opened 2019-02-04
(MODULES-8358) Add PowerShell manager to pwsh provider

glennsarti:
In preparation to add cross platform support to the PowerShell Manager,
this commit refactors the common execution code in the powershell provider into
the PuppetX::PowerShell::PowerShellManager.execute_resource method.  This
takes PowerShell code and a puppet exec resource and executes this using the
PowerShell Manager.

Previously the powershell command was a string which also contain the powershell
invocation commands.  This commit also separates the powershell command and arguments
so that we can call pwsh or powershell separately from the same PowerShell Manager.

---

Previously the pwsh provider would save a file to a temporary location and then
execute that. This is considered a legacy method.  The more modern method, which
is used in the Windows PowerShell provider, is to use the PowerShell Manager.

This commit updates the PowerShell Manager for using PowerShell Core on both
Windows and non-Windows platforms:

* The pwsh provider was modified to use the PowerShell manager if it is supported
  otherwise revert to the legacy temporary path

* The pwsh provider changed the scope of the pwsh_args and get_pwsh_command to
  public so that they can be used the spec tests.  As they're readonly properties
  it is safe to surface these methods

* Modified the PowerShell manager to use Unix Domain Sockets when on non-Windows
  platforms.  This is due to the .Net implmentation of the NamedPipesClient class
  using Domain Sockets on non-Windows platforms as opposed to Named Pipes. [1] [2] [3]
  The manager creates the domain sockets in the user's tmp directrory

  This also means the errors returned on bad/failed pipes is different e.g. EOFError
  and EConnReset.  Also many of the integration tests assume a real named pipe, so
  these tests are guarded to only run on Windows. Note that it was not possible to
  replicate the "delete named pipe" using Unix Domain sockets due to how sockets are
  different to named pipes.

* Modified many of the tests to either be platform independent by either removing the
  platform specific things e.g. remove \r\n references or specific exit codes.  Or the
  tests are modified to use specific items on specific platforms e.g. cmd.exe on Windows
  and /bin/sh on non-Windows.

* Removed the use of aliases (e.g. ps and ls) as they were causing test failures as they
  behave differently on different platforms.  Instead prefer commands like Get-Verb or
  $PSVersionTable which are consistent.

* Moved the 'when specifying a working directory' test from unit to integration as
  it actually calls the operating system for information and was causing failures when
  pwsh was not installed

* Modified the PowerShell Manager integration tests to be a shared example group and
  can then simply use 'it_behaves_like' calls to test when appropriate.  This is
  important on non-Windows when pwsh isn't installed.

* Modified the unit tests for the pwsh provider to use a mocked PowerShell Manager when
  invoked.

---

Previously the PowerShell Manager would attempt all the expected bytes from
the inbound pipe.  Normally this works fine.  However the Travis CI testing
was intermittently failing where the pipe was only reading half the pipe,
which then caused the first test to fail, and then any subsequent tests failed
because it was reading left-over data from the previous failed read.

The documentation of the IO.sysread method [1] says the argument is maxlen
which implies that this the maximum number of bytes it would return.  Note that
other methods in IO use the language "Reads at most maxlen bytes from ios"
whereas sysread uses the langauge "Reads maxlen bytes from ios"

This commit changes the pipe reader to read in chunks of bytes until all of
the expected bytes are read.  Note that this is a blocking call so if you
try to read beyond the bytes in the pipe it blocks the thread.  The value
of 8K chunk was an arbitrary number which is big enough to get most powershell
repsonse in a single IO read.  The integration tests were failing to read in
a 96K chunk (only ~36K were read)

---

- [x] Works in WSL
- [x] Works on real unix
- [x] Run PS Manager tests in unix (i.e. travis)

### DIFF
--- a/lib/puppet/provider/exec/pwsh.rb
+++ b/lib/puppet/provider/exec/pwsh.rb
@@ -17,19 +17,23 @@ Puppet::Type.type(:exec).provide :pwsh, :parent => Puppet::Provider::Exec do
   def run(command, check = false)
     @pwsh ||= get_pwsh_command
     self.fail 'pwsh could not be found' if @pwsh.nil?
-    write_script(command) do |native_path|
-      # Ideally, we could keep a handle open on the temp file in this
-      # process (to prevent TOCTOU attacks), and execute powershell
-      # with -File <path>. But powershell complains that it can't open
-      # the file for exclusive access. If we close the handle, then an
-      # attacker could modify the file before we invoke powershell. So
-      # we redirect powershell's stdin to read from the file. Current
-      # versions of Windows use per-user temp directories with strong
-      # permissions, but I'd rather not make (poor) assumptions.
-      if Puppet::Util::Platform.windows?
-        return super("cmd.exe /c \"\"#{native_path(@pwsh)}\" #{legacy_args} -Command - < \"#{native_path}\"\"", check)
-      else
-        return super("/bin/sh -c \"#{native_path(@pwsh)} #{legacy_args} -Command - < #{native_path}\"", check)
+    if PuppetX::PowerShell::PowerShellManager.supported_on_pwsh?
+      return ps_manager.execute_resource(command, resource)
+    else
+      write_script(command) do |native_path|
+        # Ideally, we could keep a handle open on the temp file in this
+        # process (to prevent TOCTOU attacks), and execute powershell
+        # with -File <path>. But powershell complains that it can't open
+        # the file for exclusive access. If we close the handle, then an
+        # attacker could modify the file before we invoke powershell. So
+        # we redirect powershell's stdin to read from the file. Current
+        # versions of Windows use per-user temp directories with strong
+        # permissions, but I'd rather not make (poor) assumptions.
+        if Puppet::Util::Platform.windows?
+          return super("cmd.exe /c \"\"#{native_path(@pwsh)}\" #{pwsh_args.join(' ')} -Command - < \"#{native_path}\"\"", check)
+        else
+          return super("/bin/sh -c \"#{native_path(@pwsh)} #{pwsh_args.join(' ')} -Command - < #{native_path}\"", check)
+        end
       end
     end
   end
@@ -41,11 +45,8 @@ Puppet::Type.type(:exec).provide :pwsh, :parent => Puppet::Provider::Exec do
     true
   end
 
-  private
-
   # Retrieves the absolute path to pwsh
   #
-  # @api private
   # @return [String] the absolute path to the found pwsh executable.  Returns nil when it does not exist
   def get_pwsh_command
     if Puppet::Util::Platform.windows?
@@ -59,10 +60,9 @@ Puppet::Type.type(:exec).provide :pwsh, :parent => Puppet::Provider::Exec do
       # ok to have 'Path' and 'PATH'
       current_env = Puppet::Util.get_environment
     end
-
     # If the resource specifies a search path use that. Otherwise use the default
     # PATH from the environment.
-    search_paths = @resource['path'].nil? ?
+    search_paths = @resource.nil? || @resource['path'].nil? ?
       current_env['PATH'] :
       resource[:path].join(File::PATH_SEPARATOR)
 
@@ -80,6 +80,21 @@ Puppet::Type.type(:exec).provide :pwsh, :parent => Puppet::Provider::Exec do
     end
   end
 
+  def pwsh_args
+    ['-NoProfile', '-NonInteractive', '-NoLogo', '-ExecutionPolicy', 'Bypass']
+  end
+
+  private
+
+  # Retrieves the PowerShell manager specific to our pwsh binary in this resource
+  #
+  # @api private
+  # @return [PuppetX::PowerShell::PowerShellManager] The PowerShell manager for this resource
+  def ps_manager
+    debug_output = Puppet::Util::Log.level == :debug
+    PuppetX::PowerShell::PowerShellManager.instance(@pwsh, pwsh_args, debug: debug_output)
+  end
+
   def write_script(content, &block)
     Tempfile.open(['puppet-pwsh', '.ps1']) do |file|
       file.puts(content)
@@ -95,9 +110,5 @@ Puppet::Type.type(:exec).provide :pwsh, :parent => Puppet::Provider::Exec do
     else
       path
     end
-  end
-
-  def legacy_args
-    '-NoProfile -NonInteractive -NoLogo -ExecutionPolicy Bypass'
   end
 end
--- a/lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
+++ b/lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
@@ -394,10 +394,18 @@ Invoke-PowerShellUserCode @params
         pipe_reader = Thread.new(@pipe) do |pipe|
           # read a Little Endian 32-bit integer for length of response
           expected_response_length = pipe.sysread(4).unpack('V').first
-          return nil if expected_response_length == 0
 
-          # reads the expected bytes as a binary string or fails
-          pipe.sysread(expected_response_length)
+          if expected_response_length == 0
+            nil
+          else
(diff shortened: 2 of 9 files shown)


lib/puppet/provider/exec/pwsh.rb
```diff
@@ -17,19 +17,45 @@
...
-      # the file for exclusive access. If we close the handle, then an
-      # attacker could modify the file before we invoke powershell. So
-      # we redirect powershell's stdin to read from the file. Current
-      # versions of Windows use per-user temp directories with strong
-      # permissions, but I'd rather not make (poor) assumptions.
-      if Puppet::Util::Platform.windows?
-        return super("cmd.exe /c \"\"#{native_path(@pwsh)}\" #{legacy_args} -Command - < \"#{native_path}\"\"", check)
-      else
-        return super("/bin/sh -c \"#{native_path(@pwsh)} #{legacy_args} -Command - < #{native_path}\"", check)
+    if PuppetX::PowerShell::PowerShellManager.supported_on_pwsh?
```
Iristyle: I assume this code is straight copied from the `powershell` provider?

If so, `pwsh` and `powershell` are probably pretty close to one another and perhaps a refactor is in order to remove the duplication.

lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
```diff
@@ -38,30 +47,53 @@ def self.supported?
...
+          # .Net implements named pipes under Linux etc. as Unix Sockets in the filesystem
+          # Paths that are rooted are not munged within C# Core.
+          # https://github.com/dotnet/corefx/blob/94e9d02ad70b2224d012ac4a66eaa1f913ae4f29/src/System.IO.Pipes/src/System/IO/Pipes/PipeStream.Unix.cs#L49-L60
+          # https://github.com/dotnet/corefx/blob/a10890f4ffe0fadf090c922578ba0e606ebdd16c/src/System.IO.Pipes/src/System/IO/Pipes/NamedPipeServerStream.Unix.cs#L44
+          # https://github.com/dotnet/corefx/blob/a10890f4ffe0fadf090c922578ba0e606ebdd16c/src/System.IO.Pipes/src/System/IO/Pipes/NamedPipeServerStream.Unix.cs#L298-L299
+          named_pipe_name = File.join(Dir.tmpdir, "#{SecureRandom.uuid}PuppetPsHost")
+          pipe_path = named_pipe_name
+        end
+        pipe_timeout = options[:pipe_timeout] || self.class.default_options[:pipe_timeout]
+        debug = options[:debug] || self.class.default_options[:debug]
```
Iristyle: Since you've already merged in the defaults, why is this necessary?

lib/puppet/provider/exec/pwsh.rb
glennsarti: Yeah potentially could make a shared function for this.

lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
glennsarti: The initialize method is public so strictly speaking I couldn't guarantee that those options would be passed in.  Should probably mark initialize as private and then I can get rid of that.

glennsarti: Derp....You can't make initialize private (It's the constructor)

glennsarti:
Looks like I have a parallel_spec/shared state issue.  I can get the tests to fail locally in parallel_spec or as a group e.g.  `rspec spec/integration`

But if I run the specific file rspec spec/integration/puppet_x/..../..._spec.rb` it's all fine.  Might need a `before(:all)` clause to clear out any instances

glennsarti:
Hrmm looks like the pipe (socket) gets into a bad state and then never recovers ....

```
1) On PowerShell Core it should behave like a PowerShellManager PuppetX::PowerShell::PowerShellManager when provided powershell commands should be able to write more than the 64k default buffer size to the managers pipe without deadlocking the Ruby parent process or breaking the pipe
     Failure/Error: expect(result[:stdout].length).to eq("#{buffer_string_96k}#{line_end}".length)
     
       expected: 98306
            got: 36493
     
       (compared using ==)
     Shared Example Group: "a PowerShellManager" called from ./spec/integration/puppet_x/puppetlabs/powershell_manager_spec.rb:810
     # ./spec/integration/puppet_x/puppetlabs/powershell_manager_spec.rb:569:in `block (4 levels) in <top (required)>'
```

glennsarti:
Figured it out.  There was a race condition _somewhere_ where the pipe reading wouldn't return all of the bytes in the pipe and needed chunking/buffering.

lib/puppet/provider/exec/pwsh.rb
glennsarti: Done!

glennsarti:
So found the issue with beaker host generator.  Fixed that and also added the ability to specify which version of powershell core to use. e.g. to use PS Core 6.1.2

```
bundle exec beaker-hostgenerator ubuntu1604-64default.a{powershell=6.1.2} | Set-Content 'hosts.yaml'
bundle exec beaker-hostgenerator windows2012r2_wmf5-64default.a{powershell=6.1.2} | Set-Content 'hosts.yaml'
```

lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
```diff
@@ -394,10 +394,18 @@ def read_streams
...
-          # reads the expected bytes as a binary string or fails
-          pipe.sysread(expected_response_length)
+          if expected_response_length == 0
+            nil
+          else
+            # reads the expected bytes as a binary string or fails
+            buffer = ""
+            # Reads in the pipe data 8K, or less, at a time
+            while (buffer.length < expected_response_length)
+              buffer << pipe.sysread([expected_response_length - buffer.length, 8192].min)
```
Iristyle: Interesting....

glennsarti: Yeah "interesting" was not the word I was using ...

Iristyle: So the problem with this is that it doesn't really solve whatever caused this problem - and it really just changes the way in which the problem may appear in the future.

What's to say that asking for 8k rather than asking for the whole thing doesn't lock the same way?

There might be a few clues in https://www.ruby-forum.com/t/sysread-changes-behavior-in-the-presence-of-threads/61176/4

It might make sense to look at the Ruby C code rather than reading the docs. If `sysread` doesn't behave the same way across platforms, it may be better to use something that's consistent.

Iristyle: This is also a really inefficient way of doing allocations in Ruby and will generate a lot of garbage.

Iristyle: Oh and not sure why the short circuit return was removed in favor of deeper branching?

glennsarti: > .. early return was removed

You can't call `return` in that context (block).  Ruby was actually throwing errors but the Thread was swallowing them `LocalJumpError: unexpected return`

glennsarti: > So the problem with this is that it doesn't really solve whatever caused this problem - and it really just changes the way in which the problem may appear in the future.

Sadly yes :-(

Iristyle: I always forget that rule about the block and should have looked up the rule again before making my comment (apologies for being in a bit of a rush when I left those comments). I think we can use `break` (which returns `nil` by default) or to be more explicit `break nil`.

https://makandracards.com/makandra/46939-ruby-a-small-summary-of-what-return-break-and-next-means-for-blocks

glennsarti: Nope.... `break` throws
`#<LocalJumpError: break from proc-closure>`

PR #249, opened 2019-02-28
(MODULES-8359) Remove non-Windows support for powershell provider

glennsarti:
Previously in MODULES-3495 cross platform support was added to the powershell
provider.  However as part of MODULES-8358 the pwsh provder was added to
achieve the same.  This means that the non-Windows platform support within the
powershell provider can be removed as it is no longer necessary.  Also note that
as this was marked as experimental, and had known issues with functioning
therefore the removal will not require a major version bump of the module.

This commit removes the support and also modifies the unit and acceptance tests
to not test the powershell provider on non-Windows platforms.


glennsarti:
This passed adhoc at [Jenkins Adhoc](https://jenkins-master-prod-1.delivery.puppetlabs.net/view/modules/view/windows/view/_adhoc/view/powershell/job/forge-windows_puppetlabs-powershell_init-cinext_nightly-skippable_adhoc/6/) at commit 
e506671.  The push to commit 68ab3dd was just to update the commit msg.  I can run through adhoc again if needed.  Otherwise this is ready for merge.

PR #250, opened 2019-03-19
(MODULES-8748) Improve pipe reading in the PowerShell Manager

glennsarti:
Previously the powershell manager was modified to read in data from the pipe
until all data was read, in 8K chunks.  However it was noted that this would be
causing more IO calls than needed due to the 8K chunks.

In this commit, the reader will attempt to read all bytes in a single call and
use the string concatenation method to reduce the number the potential number
of internal objects being created.

- [x] Add comments about why iterate


glennsarti:
@Iristyle I did some looking into potential ways to fix the sysread() issue, but they all kept coming back to, even if I knew exactly what was causing it, I wouldn't be able to solve around it.  In particular buffered IO issues, which this seems to feel like. If the underlying IO system is buffering text before it even gets into ruby or the compiled ruby layer, there's no way around it.  Particularly as this only appeared on Unix Domain Sockets (File System Based) not Named Pipes.

So instead I took the route to reduce the number of `sysread` calls to a minimum to reduce the object creation etc.  So instead of 8K chunks, the sysread call will attempt to read the entire string length from
the pipe.

---

I also did some research on the string concatenation and the general vibe I have is that using `.concat` is the fastest and most efficient method of string concatenation.

The only other method would be to allocate the entire string upfront, as we already know its expected length, and modify it as if it was a byte array.  However I can't see any ruby methods to allow me to do that besides an actual array.  And at that point I'd need to allocate an array and then convert that to a string so I'm copying the object there for sure.

---

As for the `return nil` and break.  You can't use `break` either.  It looks like the `Thread.new` is creating a proc (https://ruby-doc.org/core-2.2.0/Thread.html#method-c-new) so had to use `next` instead.

glennsarti:
I tested this by running the spec tests in a loop twenty times in a docker container

```
Linux fdb9989a9e75 4.9.125-linuxkit #1 SMP Fri Sep 7 08:20:28 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
```

lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
```diff
@@ -395,17 +395,14 @@ def read_streams
...
-            end
-            buffer
+          next nil if expected_response_length == 0
+          # reads the expected bytes as a binary string or fails
+          buffer = ""
+          # Read in the pipe data until complete
+          while (buffer.length < expected_response_length)
+            buffer.concat(pipe.sysread(expected_response_length - buffer.length))
           end
+          buffer
```
Iristyle: Ok... so I think you're actually back to a more complicated form of square one here.  The original code before your changes for 8k was:

```ruby
return nil if expected_response_length == 0

# reads the expected bytes as a binary string or fails
pipe.sysread(expected_response_length)
```

In the new code, buffer starts at length 0, so let's fill in those values for the first iteration:

```
while (0 < expected_length)
  buffer.concat(pipe.sysread(expected_length - 0))
end
```

So it seems like you might not be correctly reproing the initial failure conditions you had if this works now...

glennsarti: So as a nit, I should probably be using `do loop` not `while`.

the `concat` is accumulating the `sysread` calls. so the first call has zeros of course, but subsequent will have the buffer either full or partially full.

```
C:\Source\puppetlabs-powershell\powershell [fix-pipe-reader ≡]> irb
irb(main):001:0> buffer = ""
=> ""
irb(main):002:0> buffer.concat("Hello World")
=> "Hello World"
irb(main):003:0> buffer
=> "Hello World"
```

glennsarti: I changed the `while` to `do`

Iristyle: To follow up on this... the part that I was missing is that `sysread` reads "up to" the specified length, but no guarantees that it reads all.

So we need to keep track of what we've actually read vs what we expect to read.

glennsarti:
Added comment:
> sysread may not return all of the requested bytes due to buffering or the
> underlying IO system. Keep reading from the pipe until all the bytes are read

lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
```diff
@@ -395,17 +395,16 @@ def read_streams
...
-            end
-            buffer
+          next nil if expected_response_length == 0
+          # reads the expected bytes as a binary string or fails
+          buffer = ""
+          # sysread may not return all of the requested bytes due to buffering or the
+          # underlying IO system. Keep reading from the pipe until all the bytes are read
+          loop do
+            buffer.concat(pipe.sysread(expected_response_length - buffer.length))
+            break if buffer.length == expected_response_length
```
Iristyle: Infinite loops are always a little risky... are there other conditions that can / should exit? Should we change `==` to `>=`?

glennsarti: agreed

glennsarti:
Peending investigation into Travis CI failures

lib/puppet_x/puppetlabs/powershell/powershell_manager.rb
glennsarti: Done!

glennsarti:
I tested the new code with a 5MB payload as well and it all worked (albeit quite slow)!

PR #258, opened 2019-06-11
pdksync - FM-7993 PDK Update with pdksync_1.10.0-0-7827fc2

carabasdaniel:
FM-7993 PDK Update with pdksync_1.10.0-0-7827fc2
pdk version: `1.8.0`


.pdkignore
```diff
@@ -22,21 +22,3 @@
 /convert_report.txt
 /update_report.txt
 .DS_Store
-.project
```
glennsarti: This seems VERY odd.  Why are these all being removed?

PR #266, opened 2019-10-24
(FM-8643) Replace requires with feature confine

michaeltlombardi:
Prior to this commit the providers loaded the ruby-pwsh library
via a requires statement. When the powershell module is installed
without the dependency module, pwshlib, this requires statement
blows up as the library cannot be found.

This commit adds a ruby-pwsh feature and leverages it for the
providers.


michaeltlombardi:
@puppetlabs/modules I've updated this PR with a second implementation option. Some notes:

- To see the various implementations you'll have to review on a commit-by-commit basis.
- There really isn't a good way to generalize this pattern across modules, since the problem is we can't depend on, well, dependencies being installed
- We could make this less of a hassle for future module authors with a PDK template or helpers of some sort, perhaps
- If we feature confines we cannot recommend using the library until the resource API supports them
- This explosion is _only_ possible if modules are installed without regard for dependency chains. Should we instead let the run explode but with an explicit and clear error message that the `pwshlib` module is not installed in the environment?

michaeltlombardi:
And the third way, with an explicit failure pointing you at the right thing to do (commit incoming, had to blow away and rebuild bundle, need to test):

image

glennsarti:
Does this work in a master-agent situation?

PR #269, opened 2019-12-02
Start managing CI files

DavidS:
This requires a few changes to the pdk-templates from this PR: https://github.com/puppetlabs/pdk-templates/pull/293

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,4 @@
 ---
-os:
-  - linux
-  - osx # only tests Mac OSX edition of PowerShell Core
 dist: xenial
 language: ruby
 cache: bundler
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -1,22 +1,14 @@
 ---
 os:
   - linux
-  # OSX Only tests on the latest Puppet Gem, not the full matrix as there's no need to double up
```
glennsarti: CAn probably move this comment into .sync.yml

DavidS:
@glennsarti I don't see how this PR would disable the testing - I'll have a more in-depth look later, but would appreciate any insight you have

glennsarti:
@DavidS It looks like the stages bit is "breaking" it. There's no `CHECK` environment variable.

DavidS:
@glennsarti that seems to be a problem unrelated to this PR. Looking at a previous run from https://travis-ci.org/puppetlabs/puppetlabs-powershell/builds/620050367 the exactly same behaviour is there. Might need a separate ticket then.

glennsarti:
@DavidS Ok so looks like https://github.com/puppetlabs/puppetlabs-powershell/pull/268 broke the testing 😢

IMO that needs to be rectified before this PR can be merged.  However I'm not a primary maintainer so 🤷‍♂

DavidS:
Due to travis' propensity to multiply cells within a stage, I've now removed the main `os` key and instead put a singular `osx` cell into the acceptance stage (due to its duration). Now travis is testing on OS-X as initially planned.

We can revisit this when moving over to litmus.

DavidS:
@glennsarti are you happy with this solution?

PR #277, opened 2020-01-14
(MODULES-10389) - Safeguard powershell provider loading

michaeltlombardi:
This modifies the providers to use a confine statement to a feature found in the dependent pwshlib module, preventing Puppet runs from failing outright if the dependency is not found.


david22swan:
Updated code so that ruby-pwsh gem is no longer pointing at master

PR #286, opened 2020-02-10
Modules 10539

iakovgan:
Looks like powershell module with tag 3.0.1 brakes execution of puppet on out linux hosts, where we have no PowerShell nor pwsh.

By default ruby rescue "StandardError", but the catch is that "LoadError" is not a subset of "StandardError"
https://ruby-doc.org/core-2.5.0/Exception.html

https://tickets.puppetlabs.com/browse/MODULES-10539

### DIFF
--- a/lib/puppet/provider/exec/powershell.rb
+++ b/lib/puppet/provider/exec/powershell.rb
@@ -8,7 +8,7 @@ Puppet::Type.type(:exec).provide :powershell, :parent => Puppet::Provider::Exec 
     begin
       require 'ruby-pwsh'
       Pwsh::Manager.powershell_path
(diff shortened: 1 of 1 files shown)


michaeltlombardi:
@iakovgan thank you for this fix!

While it _would_ bandaid the problem, there's a deeper issue - this code should not be evaluating at all and instead relying on the feature confine to ensure the dependency is installed; this is only running because of the way the `commands` provider idiom is implemented.

I have filed #287 with a more thorough fix - would very much appreciate if you could test that in your environment and let me know if it works for you!

michaeltlombardi:
Superceded by #287.

PR #287, opened 2020-02-13
(MODULES-10539) Remove commands idiom from PowerShell provider

michaeltlombardi:
Prior to this commit the PowerShell exec provider used the commands idiom to
load the PowerShell executable path for use. Once the module was refactored
to rely on the ruby-pwsh gem via the puppetlabs/pwshlib module, the code for
determining the path was updated to depend on a helper function. However,
the commands method is (apparently?) evaluated during catalogue compilation,
causing the Puppet run to explode unexpectedly if the dependent module is
not available.

This commit removes the commands idiom and relies on the feature confine which
should evaluate during a Puppet run and, when that feature is not found
(because puppetlabs/pwshlib is not installed) it will report that the provider
is not functional rather than blow up the entire Puppet run.

### DIFF
--- a/lib/puppet/provider/exec/powershell.rb
+++ b/lib/puppet/provider/exec/powershell.rb
@@ -4,16 +4,6 @@ Puppet::Type.type(:exec).provide :powershell, :parent => Puppet::Provider::Exec 
   confine :operatingsystem => :windows
   confine :feature => :pwshlib
 
-  def self.powershell_path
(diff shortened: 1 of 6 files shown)


michaeltlombardi:
This PR is in draft format until the litmus acceptance test rewrites are completed and merged so the changes can be verified in acceptance and additional tests provided if necessary.

PR #299, opened 2020-04-01
Remove system-tests group from Gemfile

DavidS:
Neither beaker nor puppet-blacksmith is used anymore.


DavidS:
resetting PR for travis.

PR #304, opened 2020-05-20
(IAC-746) - Add ubuntu 20.04 support

david22swan:
IAC-746
pdk version: `1.17.0`


david22swan:
Closing as Powershell Core does not yet support Ubuntu 20.04, although it is listed as being considered in the near future.

PR #318, opened 2020-11-19
(DONOTMERGE) Extend timeout to 5m

michaeltlombardi:

### DIFF
--- a/lib/puppet/provider/exec/powershell.rb
+++ b/lib/puppet/provider/exec/powershell.rb
@@ -47,7 +47,7 @@ Puppet::Type.type(:exec).provide :powershell, :parent => Puppet::Provider::Exec 
 
   def ps_manager
     debug_output = Puppet::Util::Log.level == :debug
-    Pwsh::Manager.instance(Pwsh::Manager.powershell_path, Pwsh::Manager.powershell_args, debug: debug_output)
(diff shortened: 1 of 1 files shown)


lib/puppet/provider/exec/powershell.rb
```diff
@@ -47,7 +47,7 @@ def self.upgrade_message
 
   def ps_manager
     debug_output = Puppet::Util::Log.level == :debug
-    Pwsh::Manager.instance(Pwsh::Manager.powershell_path, Pwsh::Manager.powershell_args, debug: debug_output)
+    Pwsh::Manager.instance(Pwsh::Manager.powershell_path, Pwsh::Manager.powershell_args, debug: debug_output, pipe_timeout: 300)
```
joshcooper, line 50: what was the timeout previously?

pillarsdotnet: It defaults to 30 here: https://github.com/puppetlabs/ruby-pwsh/blob/main/lib/pwsh.rb#L37

binford2k (approved):
lgtm.

We can talk about adding a generic scripting host timeout to exec another time. And the possibility of extending the resource API to powershell providers is a thing we should consider too. But neither of those needs to wait on this.

michaeltlombardi:
Closed in favor of #321

PR #321, opened 2020-12-01
(MODULES-10722) Inherit pipe_timeout from timeout

michaeltlombardi:
Prior to this commit there was no way to modify the pipe_timeout
parameter when instantiating a PowerShell Manager instance, users
were stuck with the default defined in ruby-pwsh (30s). If this
time is insufficient to start the host process, the exec failed
with no way to extend it short of modifying the provider.

This commit updates the definition of the ps_manager by checking
to see if the timeout parameter is specified on the resource and,
if so, inheriting that to re-use for the pipe timeout.

### DIFF
--- a/lib/puppet/provider/exec/powershell.rb
+++ b/lib/puppet/provider/exec/powershell.rb
@@ -45,9 +45,9 @@ Puppet::Type.type(:exec).provide :powershell, :parent => Puppet::Provider::Exec 
     @upgrade_warning_issued = true
   end
 
-  def ps_manager
+  def ps_manager(pipe_timeout)
     debug_output = Puppet::Util::Log.level == :debug
-    Pwsh::Manager.instance(Pwsh::Manager.powershell_path, Pwsh::Manager.powershell_args, debug: debug_output)
+    Pwsh::Manager.instance(Pwsh::Manager.powershell_path, Pwsh::Manager.powershell_args, debug: debug_output, pipe_timeout: pipe_timeout)
   end
 
   def run(command, check = false)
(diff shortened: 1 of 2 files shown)


lib/puppet/provider/exec/powershell.rb
```diff
@@ -45,9 +45,13 @@ def self.upgrade_message
...
   end
 
-  def ps_manager
+  def ps_manager(resource)
     debug_output = Puppet::Util::Log.level == :debug
-    Pwsh::Manager.instance(Pwsh::Manager.powershell_path, Pwsh::Manager.powershell_args, debug: debug_output)
+    if resource[:timeout].nil?
+      Pwsh::Manager.instance(Pwsh::Manager.powershell_path, Pwsh::Manager.powershell_args, debug: debug_output)
+    else
+      Pwsh::Manager.instance(Pwsh::Manager.powershell_path, Pwsh::Manager.powershell_args, debug: debug_output, pipe_timeout: resource[:timeout])
```
glennsarti, line 53: IIRC pipe_timeout can be nil and ruby_pwsh uses it default.  So probbaly no need for the if?

glennsarti: Actually do we need the whole resource? could it just be the timeout value?

lib/puppet/provider/exec/powershell.rb
```diff
@@ -77,7 +77,7 @@ def execute_resource(powershell_code, resource)
     timeout_ms = resource[:timeout].nil? ? nil : resource[:timeout] * 1000
     environment_variables = resource[:environment].nil? ? [] : resource[:environment]
 
-    result = ps_manager.execute(powershell_code, timeout_ms, working_dir, environment_variables)
+    result = ps_manager(resource[:timeout]).execute(powershell_code, timeout_ms, working_dir, environment_variables)
```
glennsarti, line 80: Shame we have to list the timeout twice.... oh well.

sanfrancrisko (approved):

PR #358, opened 2021-08-17
(IAC-1734) - Certify Debian 11

david22swan:


david22swan:
Support for Debian 11 does not seem to have been released yet

PR #390, opened 2022-11-03
(CONT-218) - Add support for array commands

jordanbreen28:
Prior to this PR, the powershell/pwsh provider was unable to accept commands as arrays. 
This PR aims to introduce changes that will now allow the powershell and pwsh providers to accept these commands.

### DIFF
--- a/lib/puppet/provider/exec/powershell.rb
+++ b/lib/puppet/provider/exec/powershell.rb
@@ -41,7 +41,7 @@ Puppet::Type.type(:exec).provide :powershell, :parent => Puppet::Provider::Exec 
   Puppet (including 3.x), or to a Puppet version newer than 3.x and ensure you
   have .NET Framework 3.5 installed.
   UPGRADE
-
+ 
   def self.upgrade_message
     Puppet.warning POWERSHELL_MODULE_UPGRADE_MSG if !@upgrade_warning_issued
     @upgrade_warning_issued = true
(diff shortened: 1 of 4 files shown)


lib/puppet/provider/exec/powershell.rb
```diff
@@ -78,8 +78,17 @@ def execute_resource(powershell_code, resource)
     end
     timeout_ms = resource[:timeout].nil? ? nil : resource[:timeout] * 1000
     environment_variables = resource[:environment].nil? ? [] : resource[:environment]
+    
+    if powershell_code.is_a?(Array)
```
chelnak, line 82: You are on the right track here but I think this needs some more work.

I wonder if we could add some shell sanitisation to the strings passed through.

For example, if the command is an array, we could safely join the string together and if it is a string we could ensure that it is safely escaped.

There are some methods in the Shellwords library that we could make use of. However, I'm not entirely sure how they would impact commands sent to PowerShell.

Lets pair next week and explore some options together.

lib/puppet/provider/exec/powershell.rb
```diff
@@ -100,6 +108,10 @@ def checkexe(command)
   end
 
   def validatecmd(command)
+    unless command.is_a?(String) || command.is_a?(Array)
+      raise ArgumentError, _("Command must be an Array<String>, got value of class %{klass}") % { klass: command.class }
```
chelnak, line 112: I think it would be neater to use `#{}` style interpolation here. Also I think the error message needs to be expanded slightly as you are checking for a String OR an Array.

For example:

```suggestion
      raise ArgumentError, _("Command must be a String or Array<String>, got value of class #{command.class}")
```

lib/puppet/provider/exec/pwsh.rb
```diff
@@ -46,6 +46,10 @@ def checkexe(command)
   end
 
   def validatecmd(command)
+    unless command.is_a?(String) || command.is_a?(Array)
```
chelnak, line 49: Comments in powershell.rb apply here too.

lib/puppet/provider/exec/pwsh.rb
```diff
@@ -83,7 +87,16 @@ def execute_resource(powershell_code, resource)
     timeout_ms = resource[:timeout].nil? ? nil : resource[:timeout] * 1000
     environment_variables = resource[:environment].nil? ? [] : resource[:environment]
 
-    result = ps_manager(resource[:timeout]).execute(powershell_code, timeout_ms, working_dir, environment_variables)
+    if powershell_code.is_a?(Array)
```
chelnak, line 90: Comments in powershell.rb apply here too.

chelnak (changes requested):

PR #427, opened 2025-01-20
(CAT-2205): Replaced legacy fact `osfamily` with `os.family`

span786:
## Summary
Replaced legacy fact `osfamily` with `os.family`

## Additional Context
As a part of making modules 'puppet 8` compatible without deprecation warnings, we are replacing these legacy facts with the updated one.

## Checklist
- [x] 🟢 Spec tests.
- [x] 🟢 Acceptance tests.
- [ ] Manually verified. (For example `puppet apply`)


span786:
This will be covered in https://github.com/puppetlabs/puppetlabs-powershell/pull/428

